PR #25, opened 2016-08-01
406762: [xbase][compiler] UnsupportedOperationException on "return return"

LorenzoBettini:
Task-Url: https://bugs.eclipse.org/bugs/show_bug.cgi?id=406762


LorenzoBettini:
Build runs fine http://services.typefox.io/open-source/jenkins/job/xtext-extras/job/lb_bug_406762/

please review

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTests2.xtend
```diff
@@ -1856,4 +1856,16 @@ class CompilerTests2 extends AbstractOutputComparingCompilerTests {
...
 
+	@Test def void test406762_NestedReturn() {
+		// this requires argument preparation
+		'''
+			{
+				return return 0
+			}
+		'''.compilesTo('''
+			return 0;
+			return error - couldn't compile nested return;
```
szarnekow: The compiled code should not contain syntax errors.

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
```diff
@@ -1128,4 +1128,12 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
 			}
 		'''.expression.assertNoErrors
 	}
+
+	@Test def void testInvalidNestedReturnBug406762() {
+		'''
+			{
+				return return 0
```
szarnekow: Please add more tests for cases like

- return { return }
- return throw ...
- return if (true) return 0 else return 1;

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
```diff
@@ -401,24 +401,29 @@ public void checkTypeReferenceIsNotVoid(XCasePart expression) {
 	
 	@Check
 	public void checkReturn(XReturnExpression expr) {
+		XExpression returnedExpression = expr.getExpression();
```
szarnekow: My gut feeling is that this is more a type system issue and should be handled there. When the type for a return expression is computed, the state likely contains information that we are in the context of another return expression.

szarnekow:
Hi Lorenzo, thank you for the path. Please add more test cases as outlined and please also check whether another location is more appropriate for the fix. I'd think the type computation does already have enough context to validate a return expression.

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTests2.xtend
LorenzoBettini: I saw the same pattern in the XbaseCompiler. You're suggesting to generate that part in a Java comment?

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
LorenzoBettini: Right! In all of the above cases we should get a validation error, right? Or are there situations when a nested return is considered as valid?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
LorenzoBettini: OK, I'll inspect that.

LorenzoBettini:
@szarnekow could you please have a look at the modifications? I think I addressed all your suggestions. As far as I understand, that is the way to use typing information to tell whether we're in a nested return

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
svenefftinge: return throw is technically ok and there are situations where you want to allow it.
So I think we shouldn't validate that, or at least only if there is no branch that evaluates to an actual value.

E.g.
return if (true) throw else throw //invalid
return if (baz) throw else 42 // ok

svenefftinge: Also "throw return .. " is invalid.

LorenzoBettini:
@svenefftinge I've updated the PR according to your comments:

return if (true) throw else throw //invalid
return if (baz) throw else 42 // ok
throw return // invalid

I generalized the check in the type computer; in particular, the one that was checking for nested return also implements the check of a throw inside the two branches of an if.

The message is "Invalid return's expression"; it becomes more specific, "Invalid nested return." only in the case of a direct "return return".

The build fails, but that is due to https://github.com/eclipse/xtext-extras/issues/63

meysholdt:
#63 is fixed, please update your branch

LorenzoBettini:
Thanks @meysholdt now the build runs fine. :)

LorenzoBettini:
@svenefftinge can I merge?

svenefftinge:
+1

PR #28, opened 2016-08-02
Integration Test for Core Issue 65

cdietrich:
Integ test for https://github.com/eclipse/xtext-core/issues/65


cdietrich:
p.s. i will undo the second commit before the actual merge

PR #31, opened 2016-08-07
[40006] fixed issue with integerrange list literals in for loop. 

cdietrich:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=440006

Signed-off-by: Christian Dietrich christian.dietrich@itemis.de

### DIFF
--- a/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
+++ b/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
@@ -1246,6 +1246,105 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
 			}
 		'''.expression.assertNoErrors
 	}
+	
+	@Test def void testRangeLiteralInForLoopBug440006_01() {
+		'''
+			for (int i : #[1..2]) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XBINARY_OPERATION, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_02() {
+		'''
+			for (Integer i : #[1..2]) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XBINARY_OPERATION, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_03() {
+		'''
+			for (Number i : #[1..2]) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XBINARY_OPERATION, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_04() {
+		'''
+			for (i : #[1..2]) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_05() {
+		'''
+			for (IntegerRange i : #[1..2]) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_06() {
+		'''
+			for (Iterable<Integer> i : #[1..2]) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_07() {
+		'''
+			for (int i : #[1..2].flatten) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_01b() {
+		'''
+			for (int i : #{1..2}) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XSET_LITERAL, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_02b() {
+		'''
+			for (Integer i : #{1..2}) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XSET_LITERAL, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+
+	@Test def void testRangeLiteralInForLoopBug440006_03b() {
+		'''
+			for (Number i : #{1..2}) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XSET_LITERAL, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_04b() {
+		'''
+			for (i : #{1..2}) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_05b() {
+		'''
+			for (IntegerRange i : #{1..2}) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_06b() {
+		'''
+			for (Iterable<Integer> i : #{1..2}) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_07b() {
+		'''
+			for (int i : #{1..2}.flatten) {
+			}
+		'''.expression.assertNoErrors
+	}
 
 	def private assertNestedReturn(CharSequence input, EClass objectType) {
 		input.expression.assertError(
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/CollectionLiteralsTypeComputer.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/CollectionLiteralsTypeComputer.java
@@ -176,7 +176,7 @@ public class CollectionLiteralsTypeComputer extends AbstractTypeComputer {
 
 	protected void computeArrayLiteralType(XListLiteral literal, LightweightTypeReference expectedArrayType, ITypeExpectation expectation,
 			ITypeComputationState state) {
-		LightweightTypeReference elementTypeExpectation = expectedArrayType.getComponentType();
+		LightweightTypeReference elementTypeExpectation = getElementOrComponentType(expectedArrayType, state);
 		int allFlags = 0;
 		for(XExpression element: literal.getElements()) {
 			ITypeComputationResult elementTypeResult = computeTypes(element, elementTypeExpectation, state);
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/ElementOrComponentTypeComputer.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/ElementOrComponentTypeComputer.java
@@ -13,6 +13,7 @@ import org.eclipse.xtext.common.types.JvmGenericType;
 import org.eclipse.xtext.common.types.JvmTypeParameter;
 import org.eclipse.xtext.xbase.typesystem.references.AnyTypeReference;
 import org.eclipse.xtext.xbase.typesystem.references.ArrayTypeReference;
+import org.eclipse.xtext.xbase.typesystem.references.CompoundTypeReference;
 import org.eclipse.xtext.xbase.typesystem.references.ITypeReferenceOwner;
 import org.eclipse.xtext.xbase.typesystem.references.LightweightMergedBoundTypeArgument;
 import org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReference;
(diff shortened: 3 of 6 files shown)


cdietrich:
please review

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
```diff
@@ -1128,4 +1128,53 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
...
+		'''
+			for (i : #[1..2]) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_05() {
+		'''
+			for (IntegerRange i : #[1..2]) {
+			}
```
meysholdt: this is OK.

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
```diff
@@ -1128,4 +1128,53 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
...
+		'''
+			for (Integer i : #[1..2]) {
+			}
+		'''.expression.assertError(XbasePackage.Literals.XBINARY_OPERATION, IssueCodes.INCOMPATIBLE_TYPES)
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_03() {
+		'''
+			for (Number i : #[1..2]) {
+			}
```
meysholdt: This is a single for-loop iterating over something two-dimensional. I'm not in favor of supporting this because it's hiding potentially hides user's mistakes and we don't need any sugar here since it's a corner case. If somebody wants to iterate over `#[1..2]` the correct way to do so is to use two nested loops:

    for (range : #[1..2]) {
        for (Number i: range) {
            println(i)
        }
    }

It may be more illustrative when the array has one than more item:

    for (range : #[1..2, 5..9, 12..13]) {
        for (Number i: range) {
            println(i)
        }
    }

cdietrich: this is why i give a error dont i?

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
```diff
@@ -1128,4 +1128,53 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
...
+		'''
+			for (IntegerRange i : #[1..2]) {
+			}
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_06() {
+		'''
+			for (Iterable<Integer> i : #[1..2]) {
+			}
```
meysholdt: OK

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
```diff
@@ -1128,4 +1128,53 @@ class XbaseValidationTest extends AbstractXbaseTestCase {
...
+		'''.expression.assertNoErrors
+	}
+
+	@Test def void testRangeLiteralInForLoopBug440006_07() {
+		'''
+			for (int i : #[1..2].flatten) {
+			}
```
meysholdt: OK

org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/validation/XbaseValidationTest.xtend
meysholdt: true, I overlooked that one, sorry

cdietrich:
so you think it is ok to merge?

meysholdt:
not yet, I'm still testing a few things

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/CollectionLiteralsTypeComputer.java
```diff
@@ -176,7 +177,17 @@ protected void computeType(XListLiteral literal, JvmGenericType listType, ITypeE
 
 	protected void computeArrayLiteralType(XListLiteral literal, LightweightTypeReference expectedArrayType, ITypeExpectation expectation,
 			ITypeComputationState state) {
-		LightweightTypeReference elementTypeExpectation = expectedArrayType.getComponentType();
+		LightweightTypeReference elementTypeExpectation = null;
+		if (expectedArrayType instanceof CompoundTypeReference) {
+			elementTypeExpectation = new CompoundTypeReference(expectedArrayType.getOwner(), expectedArrayType.isSynonym());
```
meysholdt: please use org.eclipse.xtext.xbase.typesystem.references.ITypeReferenceOwner.newCompoundTypeReference(boolean) here

meysholdt:
besides using `org.eclipse.xtext.xbase.typesystem.references.ITypeReferenceOwner.newCompoundTypeReference(boolean)` instead of the invoking the constructor of `CompoundType` directly, the change looks fine to me.

cdietrich:
is `state.getReferenceOwner().newCompoundTypeReference(expectedArrayType.isSynonym());` ok

or `expectedArrayType.getOwner()...`

meysholdt:
`state.getReferenceOwner()` as it is more specific to the current context

meysholdt:
but don't merge this yet, somebody else needs to review this

cdietrich:
ok

akosyakov:
Is not it an issue for set literals as well?

```xtend
for (int i : #{1 .. 2}) {
}
```

cdietrich:
well yes but that seems to be a different problem

```
java.lang.UnsupportedOperationException: Missing implementation for int[] | Iterable<? extends Integer> | Integer[]
	at org.eclipse.xtext.xbase.typesystem.references.TypeReferenceVisitorWithResult.doVisitTypeReference(TypeReferenceVisitorWithResult.java:17)
	at org.eclipse.xtext.xbase.typesystem.references.TypeReferenceVisitorWithResult.doVisitCompoundTypeReference(TypeReferenceVisitorWithResult.java:29)
	at org.eclipse.xtext.xbase.typesystem.references.TypeReferenceVisitorWithResult.doVisitSynonymTypeReference(TypeReferenceVisitorWithResult.java:37)
	at org.eclipse.xtext.xbase.typesystem.references.CompoundTypeReference.accept(CompoundTypeReference.java:361)
```

cdietrich:
@akosyakov can you file a separate ticket for that

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/CollectionLiteralsTypeComputer.java
```diff
@@ -176,7 +177,17 @@ protected void computeType(XListLiteral literal, JvmGenericType listType, ITypeE
 
 	protected void computeArrayLiteralType(XListLiteral literal, LightweightTypeReference expectedArrayType, ITypeExpectation expectation,
 			ITypeComputationState state) {
-		LightweightTypeReference elementTypeExpectation = expectedArrayType.getComponentType();
+		LightweightTypeReference elementTypeExpectation = null;
```
akosyakov: Would not be calling getElementOrComponentType(expectedArrayType, state) enough?

```java
LightweightTypeReference elementTypeExpectation = getElementOrComponentType(expectedArrayType, state);
```

in ElementOrComponentTypeComputer we can override doVisitCompoundTypeReference(CompoundTypeReference):
```java
if (reference.isArray()) {
    for (LightweightTypeReference componentType : reference.getMultiTypeComponents()) {
        if (componentType.isArray()) {
            return componentType.getComponentType();
        }
    }
}
return super.doVisitCompoundTypeReference(reference);
```

Such fix will cover set literals as well.

akosyakov:
@cdietrich i think they are related

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/CollectionLiteralsTypeComputer.java
akosyakov: Could it be also that CompoundTypeReference.getComponentType for the case when CompoundTypeReference.isArray returns true is bogus?

cdietrich: i have to give that a try. but do we really want to have the first array candidate?

akosyakov: @cdietrich good question, as far as i see at least in the case of a synonym compound reference with array components it should work

cdietrich: @akosyakov  yes that is working. i will look if this destroys something and add some tests

cdietrich: it actually does not cover sets yet. have to digg for that

cdietrich:
@akosyakov @meysholdt now it is work for both sets and lists. can you have a look again?

akosyakov:
@meysholdt @cdietrich  looks good to me, but:

1. Should we have tests for for-loops inside rich-strings for Xtend?

2. I still wonder whether we can implement CompoundTypeReference.getComponentType that it looks symmetrically to CompoundTypeReference.isArray. There are places where we call isArray on LightweightTypeReference and then getComponentType, if it happens to be a compound type reference they could fail.

3. I've noticed with java 8: #[1 .. 2].forEach[int i | ] generates a bogus java code. But it seems to be unrelated to the current issue.

akosyakov:
@cdietrich I think it is fine merge the second point can be discussed and solved later

akosyakov:
https://github.com/eclipse/xtext/issues/2356 - for the third point

cdietrich:
ok ill update the pullreq for (2)

cdietrich:
added tests for (1)

spoenemann:
What's the state here? Can it be merged now?

cdietrich:
no i think @svenefftinge should do a review

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/references/CompoundTypeReference.java
```diff
@@ -113,6 +113,15 @@ public boolean isArray() {
 	}
 	
 	@Override
+	public LightweightTypeReference getComponentType() {
+		for(LightweightTypeReference component: expose(components)) {
```
svenefftinge: since we don't expose the collection here, a simple components!=null would be a bit straight and faster.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/references/CompoundTypeReference.java
```diff
@@ -113,6 +113,15 @@ public boolean isArray() {
...
+			if (component.isArray())
+				return component.getComponentType();
+		}
+		throw new IllegalStateException("CompoundTypeReference is an array but none of its components is an array.");
```
svenefftinge: I don't see a test for this#isArray(). I'd prefer to simply cal super and use the general contract to return null.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/ElementOrComponentTypeComputer.java
```diff
@@ -90,6 +91,18 @@ protected LightweightTypeReference doVisitUnboundTypeReference(UnboundTypeRefere
 		return null;
 	}
 	
+	@Override
+	protected LightweightTypeReference doVisitCompoundTypeReference(CompoundTypeReference reference) {
+		if (reference.isArray()) {
+		    for (LightweightTypeReference componentType : reference.getMultiTypeComponents()) {
```
svenefftinge: seems like you could simply call getComponentType on the compound reference, since you've implemented it below.

cdietrich:
updated pull req implementing review feedback

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/ElementOrComponentTypeComputer.java
```diff
@@ -90,6 +91,14 @@ protected LightweightTypeReference doVisitUnboundTypeReference(UnboundTypeRefere
 		return null;
 	}
 	
+	@Override
+	protected LightweightTypeReference doVisitCompoundTypeReference(CompoundTypeReference reference) {
+		if (reference.isArray()) {
+			return reference.getComponentType();
+		}
+		return super.doVisitCompoundTypeReference(reference);
```
svenefftinge: sorry, I overlooked this in my first review.
I think we shouldn't check for reference.isArray() but just delegate to reference.getComponentType() in all cases.

svenefftinge:
+1 Thank you

Issue #32, opened 2016-08-10
 [Junit4Fragment] Exported test package becomes unneccessary public API


kthoms:
Same issue as https://github.com/eclipse/xtext-core/issues/67, but for the old generator fragment.

cdietrich:
@kthoms what is left open for this issue?

kthoms:
won't be fixed

PR #33, opened 2016-08-10
[#32] Mark exported test package as internal

kthoms:


svenefftinge:
I'd rather not change such things in maintenance code, at least if there is no actual problems for clients.

kthoms:
So the change is OK, but we simply don't change the manifest of existing test plugins? That would be OK.

spoenemann:
Sven's point was that we shouldn't change anything in the old generator unless there are clients who really depend on it.

PR #41, opened 2016-08-17
[381031] Detect recursive calls of private methods

kthoms:
Covered by XtendValidationTest, see https://github.com/eclipse/xtext-xtend/pull/27

Signed-off-by: Karsten Thoms karsten.thoms@itemis.de

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
@@ -27,6 +27,8 @@ import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EPackage;
 import org.eclipse.emf.ecore.EReference;
 import org.eclipse.emf.ecore.EStructuralFeature;
+import org.eclipse.emf.ecore.EStructuralFeature.Setting;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.xtext.EcoreUtil2;
 import org.eclipse.xtext.common.types.JvmConstructor;
 import org.eclipse.xtext.common.types.JvmDeclaredType;
@@ -46,6 +48,7 @@ import org.eclipse.xtext.common.types.JvmType;
 import org.eclipse.xtext.common.types.JvmTypeConstraint;
 import org.eclipse.xtext.common.types.JvmTypeParameter;
 import org.eclipse.xtext.common.types.JvmTypeReference;
+import org.eclipse.xtext.common.types.JvmVisibility;
 import org.eclipse.xtext.common.types.JvmVoid;
 import org.eclipse.xtext.common.types.TypesPackage;
 import org.eclipse.xtext.common.types.util.DeprecationUtil;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=381031

meysholdt:
impl looks ok. Can you please add some test cases?

meysholdt:
Sorry, found the test cases.

meysholdt:
We should strive towards having tests in the same repo as the code-under-test.

The test for this PR depends on a JvmOperation but that shouldn't be a reason to have the test in the Xtend repo. See 'org.eclipse.xtext.xbase.tests.jvmmodel.SimpleJvmModelInferrer' which is being used by `org.eclipse.xtext.xbase.tests.jvmmodel.AbstractJvmModelTest`

Based on this infrastructure there already is a validation test: `org.eclipse.xtext.xbase.tests.jvmmodel.CheckExceptionValidationTest`

Could you move the test to this PR and repo, possibly by using the helpers mentioned above?

kthoms:
🆗

kthoms:
I am undecided. In Xbase there is no validation rule for this case, the isLocallyUsed method is only used in other contexts (checkLocalUsageOfDeclared, checkLocalUsageOfSwitchParameter). The check for unused private members is in Xtend only.

We could move the code that I added to XbaseValidator to XtendValidator and override the isLocallyUsed method. Then check and tests would be in the same repo. On the other hand, the code is independent from Xtend.

Since there is no check in Xbase itself, I do not see how to add a test for this. The only chance I see is to subclass XbaseValidator just for testing purpose to check the method isLocallyUsed with a JvmOperation. Or have a test class in the same package as XbaseValidator to test the protected method, which seems not common.

So which way to go?

kthoms:
@meysholdt What do you think? Should I move the code to Xtend or do you see a chance to actually write a test in Xbase?
I think it is desirable to have the code under test and the test code in the same repo, but I think in this case this is just too complicated to achieve.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
```diff
@@ -1346,7 +1348,29 @@ public void checkPrimitiveComparedToNull(XBinaryOperation binaryOperation) {
...
+			// for private members it has to be checked if all usages are within the operation
+			EObject targetSourceElem = associations.getPrimarySourceElement(target);
+			long featureCallsWithinTarget = usages.stream()
+				// filter all feature calls to target
+				.map(s -> s.getEObject())
+				.filter(it -> it instanceof XFeatureCall)
+				.map(XFeatureCall.class::cast)
+				.filter(fc->fc.getFeature()==target)
+				// filter all feature calls which are contained in the same operation
+				.filter(fc->Iterables.contains(EcoreUtil2.getAllContainers(fc), targetSourceElem))
```
spoenemann: `EcoreUtil.isAncestor(targetSourceElem, fc)`

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
```diff
@@ -1346,7 +1348,29 @@ public void checkPrimitiveComparedToNull(XBinaryOperation binaryOperation) {
...
+				.count()
+				;
+			// if not all usages are recursive calls return true
+			return usages.size() != featureCallsWithinTarget;
```
spoenemann: The Java8 style stream code above always traverses all elements. A simple for loop would be more efficient, since then you can early exit with `return true` as soon as you find an element that is not contained in the operation.

svenefftinge:
@spoenemann 
Yes, but long term we should target doing more "local" changes. 
Especially for the xbase case, where we so far reuse Xtend as a testlanguage, it might be worth to have a dedicated testlanguage in extras.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/XbaseValidator.java
kthoms: The loop would have to iterate over all elements anyway, since from all "usages" only those whose EObject is an XFeatureCall have to be considered.
`.filter(it -> it instanceof XFeatureCall)`
Within the loop there is no chance for an early exit here, since first all usages have to be considered to decide how many of these are recursive calls within the same operation.
So, I could rephrase the logic to plain old Java, but the logic would be the same.
Please advice if I missed something.

spoenemann:
@kthoms 
> The loop would have to iterate over all elements anyway, since from all "usages" only those whose EObject is an XFeatureCall have to be considered.
> `.filter(it -> it instanceof XFeatureCall)`
>Within the loop there is no chance for an early exit here, since first all usages have to be considered to decide how many of these are recursive calls within the same operation.
> So, I could rephrase the logic to plain old Java, but the logic would be the same.
> Please advice if I missed something.

What I meant is something like this
```
for (s : usages) {
    if (s.EObject instanceof FeatureCall) {
        val fc = s.EObject as FeatureCall
        if (fc.feature != target || !EcoreUtil.isAncestor(targetSourceElem, fc))
            return false;
    } else
        return false;
}
return true;
```
That should do the same as your code.

kthoms:
Updated. Please review.

spoenemann:
The Xtend test failure is probably caused by this change:
http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/master/148/testReport/junit/org.eclipse.xtend.core.tests.validation/NestedClassValidationTest/test_06/

@kthoms could you have a look?

cdietrich:
yes it is this changes @spoenemann @kthoms

kthoms:
PR https://github.com/eclipse/xtext-extras/pull/79 should fix the issue.

PR #42, opened 2016-08-18
[492072] Optimized compiler

kthoms:
Replacement for https://github.com/eclipse/xtext/pull/1043 after repository split

avoid generation of unnecessary line for getter result
see https://bugs.eclipse.org/bugs/show_bug.cgi?id=492072

Signed-off-by: Karsten Thoms karsten.thoms@itemis.de

### DIFF
--- a/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTest.java
+++ b/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTest.java
@@ -30,8 +30,7 @@ public class CompilerTest extends AbstractOutputComparingCompilerTests {
 				"\n" + 
 				"try {\n" + 
 				"  Class<?> clazz = Class.forName(\"java.lang.String\");\n" + 
-				"  Class<?> _superclass = clazz.getSuperclass();\n" + 
-				"  org.eclipse.xtext.xbase.lib.InputOutput.<Class<?>>println(_superclass);\n" + 
+				"  org.eclipse.xtext.xbase.lib.InputOutput.<Class<?>>println(clazz.getSuperclass());\n" + 
 				"} catch (Throwable _e) {\n" + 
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=492072

spoenemann:
Looks good to me, but this needs a few more reviews.

The solution addresses two particular cases, feature call arguments and assignment RHS. I wonder whether this can be generalized further.

svenefftinge:
Yes, agree with @spoenemann, I would like to simplify it and make it more general.

kthoms:
@svenefftinge @spoenemann Could you be more specific, please? Are there more concrete examples where an optimization would be valid?
Of course it is good if we could generalize it a bit more, but I would like to know where this should be possible without risking a break.
Otherwise I propose that we take this optimization now and if we find that there is more to do we should open another issue.

spoenemann:
What I'm missing here is an analysis of the cases where a synthetic variable is not necessary. Your change considers feature call arguments and assignment value. Why not also for other kinds of containers? Or putting it the other way, could we simply return `false` at the end of the `if (expr instanceof XAbstractFeatureCall)` block and explicitly handle only those cases where the synthetic variable is required?

kthoms:
> Or putting it the other way, could we simply return false at the end of the if (expr instanceof XAbstractFeatureCall) block and explicitly handle only those cases where the synthetic variable is required?

I don't think that it is valid to return just `false` there. And the previous existing code already handles for other cases either if a synthetic variable is necessary or not explicitly. So rewriting the code for only positive cases could be quite some work.

What I've done so far is to read generated Xtend code for quite some cases to find other examples where a synthetic variable is not required. I did not find more concrete examples where this could be answered with "no".

Again, I propose that if there is more concrete evidence for other cases then this could be handled with another issue. In other case I would prefer not to invest too much more time on this issue and leave it as is.

spoenemann:
Ok, then 👍 from me.

PR #47, opened 2016-08-31
46: Make setting Java version in compiler tests easier

### LINKED ISSUE #46: Make setting Java version in compiler tests easier
I'm about to contribute a PR for making setting the Java version easier when using CompilationTestHelper and OnTheFlyJavaCompiler2. This will set both the Java version for the DSL compiler (e.g., Xbase) and for the Java compiler.

LorenzoBettini:
Closes #46

Task-Url: #46


LorenzoBettini:
Build runs fine.
Please review, making sure to have a look at https://github.com/eclipse/xtext-extras/pull/47/commits/51ca6352c8ef09be6379b8d33a5999bcdbaf4a96

LorenzoBettini:
@svenefftinge could you please have a look? You're the author of the original class

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
```diff
@@ -147,6 +150,17 @@ public void setJavaCompilerClassPath(ClassLoader classLoader) {
...
+
+	/**
+	 * Sets the Java version both for the DSL generator, for example, Xbase
+	 * compiler, and for the Java compiler.
+	 * 
+	 * @since 2.11
+	 */
+	public void setJavaVersion(JavaVersion javaVersion) {
+		this.javaCompiler.setJavaVersion(javaVersion);
+		generatorConfigProvider.get(null).setJavaSourceVersion(javaVersion);
```
svenefftinge: calling the provider with null looks odd. Looking at the implementation it will return a fresh GeneratorConfig with no association.

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
```diff
@@ -212,6 +226,16 @@ public void compile(Iterable<? extends CharSequence> sources, IAcceptor<Result>
 	public void compile(final ResourceSet resourceSet, IAcceptor<Result> acceptor) {
 		try {
 			List<Resource> resourcesToCheck = newArrayList(resourceSet.getResources());
+			if (generatorConfigProvider instanceof GeneratorConfigProvider) {
+				GeneratorConfigProvider configProvider = (GeneratorConfigProvider) generatorConfigProvider;
+				GeneratorConfig config = generatorConfigProvider.get(null);
```
svenefftinge: I'd prefer a constructor call to GeneratorConfig, instead of call the provider with null

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
LorenzoBettini: @svenefftinge that looks odd to me as well, but I just mimicked what is already being done in the existing code: https://github.com/eclipse/xtext-extras/blob/7e358229d6f5cee004ab9cb0389af49b95a230ed/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java#L150

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
LorenzoBettini: @svenefftinge but if someone provides a custom implementation of GeneratorConfigProvider that would not be used (I also added a test for that https://github.com/eclipse/xtext-extras/pull/47/files#diff-c477e770a10cf16ed4e0ec0c77586b3bR161). I would have wanted to pass the resourceSet to the get method, but get() takes an EObject... the alternative would have been to pass the first EObject of the first resourc ein the resourceSet.

LorenzoBettini:
@svenefftinge thanks for the review. Please see my comments. In general it was not clear to me how to use the GeneratorConfigProvider correctly. If you think we should create a GeneratorConfig manually I'm OK with that.

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
svenefftinge: There it is done to retrieve the default value. But here you are setting the java version on an object that is not referenced anywhere, no?

LorenzoBettini: Right! I'll remove that.

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/CompilationTestHelper.java
LorenzoBettini: OK, then I'll update the PR by removing the line you mentioned above, and then I'd merge it.

Issue #48, opened 2016-09-07
[xbase] Getter/setter body uncompilable when field name is a keyword


kthoms:
To reproduce the issue use the Domainmodel example and add to the grammar:

```
ValidID:
	ID | KEYWORD
;

KEYWORD: 'package' | 'entity';
```

This will allow to use `package` and `entity` as valid identifiers for e.g. property names. The following DSL sample would be valid:
```
package test {
	entity Delivery {
		weight  : int
		package : boolean
	}
}
```

However, it will produce invalid Java code:

![screenshot 1](https://cloud.githubusercontent.com/assets/265597/18304838/f999fc88-74e4-11e6-89f4-573beca669a2.png)

While the field name is suffixed with `_` to `package_`, the getter/setter bodies do not make use of this escaping, leading to compile errors.

cdietrich:
Fixed in Xtext 2.11

PR #49, opened 2016-09-07
[#48] Escape names in getter/setter when in conflict with Java keywords

kthoms:


cdietrich:
looks good 👍

PR #59, opened 2016-09-28
[501975] Null guard for switch statement for all non-primitive types

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=501975

LorenzoBettini:
Looks good to me!
I guess this will trigger some Xtend code regeneration but that will be handled later, when this gets into nightly, right?

kthoms:
Yes, some Xtend code will compile differently then and should be checked in later. I love checked-in generated code. Not.

LorenzoBettini:
@kthoms Unfortunately, this change makes two tests fail: http://services.typefox.io/open-source/jenkins/job/xtext-extras/job/master/59/ could you please fix them?

kthoms:
I doubt that it was caused by this change, it must be some upstream change related to the serializer or formatter.
I recognized this problem also and fixed it with https://github.com/eclipse/xtext-extras/pull/60
Please review that one.

LorenzoBettini:
mh... that's strange: the commit history https://github.com/eclipse/xtext-extras/commits/master shows that after 23 Sept only this PR has been merged, and master on 23 Sept had no failures, http://services.typefox.io/open-source/jenkins/job/xtext-extras/job/master/58/ while master after merging this PR has failed tests http://services.typefox.io/open-source/jenkins/job/xtext-extras/job/master/59/

don't you experience these failed tests in your Eclipse?

kthoms:
I experience them and I fixed the SerializerTest in https://github.com/eclipse/xtext-extras/pull/60. But I do not see how this could be related to the change here. This only affects the code compiled by Xtend. Since no code that was compiled with Xtend was checked in so far, this cannot cause the fail of the test.

I just think that the failing tests are unrelated to this issue and the PR.

Please review the other PR, this will fix the tests. As discussed with Moritz on Slack, it is OK to change the expectation here.

LorenzoBettini:
So you say that probably something in xtext-core changed in the meantime and this caused these failing tests, right?

kthoms:
Yes, I think so.

PR #60, opened 2016-09-28
Fixed test

kthoms:
- Expected serialized string adjusted to actual result from the
serializer.
- Extended test to check that the original AST is equal to the one that
gets parsed from the serialized expression.


LorenzoBettini:
it was discussed with @meysholdt on Slack that it is OK to merge this.
It should also fix these failing tests: http://services.typefox.io/open-source/jenkins/job/xtext-extras/job/master/59/

PR #65, opened 2016-10-13
Created DefaultXbaseIdeModule

spoenemann:
See eclipse/xtext-core#127 and eclipse/xtext-web#6.


spoenemann:
Depends on eclipse/xtext-core#143.

PR #69, opened 2016-11-03
406762: check if the return is contained in a throw

LorenzoBettini:
It is necessary to also check that, otherwise a return inside a method
that returns a Throwable would be marked as an error.

See https://github.com/eclipse/xtext-xtend/issues/68

Task-Url: https://bugs.eclipse.org/bugs/show_bug.cgi?id=406762

### DIFF
--- a/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTests2.xtend
+++ b/org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/compiler/CompilerTests2.xtend
@@ -2042,4 +2042,55 @@ class CompilerTests2 extends AbstractOutputComparingCompilerTests {
 		''', false)
 	}
 
+	@Test def void test406762_ValidReturnInLambdaContainedInThrow() {
+		'''
+			throw {
+				val foo = [|return "foo"]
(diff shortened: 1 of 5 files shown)


LorenzoBettini:
Please review, this should fix https://github.com/eclipse/xtext-xtend/issues/68 and it does not break the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=406762

I created an Xtend test for reproducing the bug in https://github.com/eclipse/xtext-xtend/issues/68 and it works now (I'll create a separate PR for Xtend with the added test once this PR has been approved).

spoenemann:
I don't quite understand how the fix works. Doesn't `EcoreUtil2.getContainerOfType(object, XThrowExpression.class)` yield something for the test cases you provided? In those examples `return` expression is indirectly contained in a `throw` expression.

LorenzoBettini:
The inspection of the container is done only if there's a Throwable expectation. The test cases I added show that the lambda code is considered valid : since there's no Throwable expectation, then the container is not checked. I added the test cases since @svenefftinge thought that this fix would have considered these lambda expressions as invalid. 

The test cases are not related to show that the modification in this pr fixes the Xtend bug, but to show that it doesn't cause other regression.

spoenemann:
Ok, then 👍 from me

PR #80, opened 2016-12-13
Fixed bug in early-exit validation

lievenlemiengre:
Added support for switch expressions to the early-exit computer.

currently this works:
`def String foo() {
	while(true) {
		if(false) return "test"
	}
}`

but this fails:
`def String bar() {
	while(true) {
		switch false {
			case true: return "test" 
		}
	}
}`

with the message "Type mismatch: cannot convert from void to String", this PR fixes this problem

Signed-off-by: Lieven Lemiengre <lieven.lemiengre@sigasi.com>


svenefftinge:
Thank you!

PR #91, opened 2016-12-24
[492072] Optimize FeatureCallCompiler

kthoms:
This PR will tweak the FeatureCallCompiler to avoid declaration of unnecessary variables. It will make the compiled code more compact and often more readable. Test expectations are adjusted.

### DIFF
--- a/org.eclipse.xtext.purexbase.tests/src/org/eclipse/xtext/purexbase/test/CompilerTest.xtend
+++ b/org.eclipse.xtext.purexbase.tests/src/org/eclipse/xtext/purexbase/test/CompilerTest.xtend
@@ -698,12 +698,10 @@ class CompilerTest {
 			    "string".length();
 			    final Integer i = Integer.valueOf("string".length());
 			    /* i; */
-			    int _length = "string".length();
-			    Integer.valueOf(_length).toString();
+			    Integer.valueOf("string".length()).toString();
 			    "string".toCharArray();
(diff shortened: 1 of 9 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=492072

svenefftinge:
These 'improvements' to the compiler indeed make the generated code smaller. But it harms debuggability. 
The design to statementizes expression and introduce synthetic variables, was not made because we didn't know how to do it differently, but because we wanted to make functional code, which means a lot of things happening in one expression, easier debuggable.

So far with xtend (or xbase) you could write code in a very functional way without suffering from the worse debuggability, because you would get intermediate results through the statementisized expressions. Your changes removed this feature. 

It would be nice to have a compiler flag to switch between the two flavours.

holgerschill:
@svenefftinge Karsten and I already discussed that and indeed this would make sense. Thanks for the feedback. This change will fix the build again. After that we should introduce the flag and introduce the more verbose variant as the default. Is that ok for you?

cdietrich:
@holgerschill doesn't that already work?

holgerschill:
So far as I know not for Xtend expressions natively. Only when you switch to Java code.

svenefftinge:
Ok, maybe even the more compact code could be the default then.

PR #105, opened 2017-01-16
[460455] Avoid import of inner classes

kthoms:
A usage is local when the declared type is withing the same resource as
the usage context

Fixes the first issue described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=460455


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=460455

PR #106, opened 2017-01-17
Moved some long running tests to the longrunning source folder

spoenemann:
The xtext-extras build takes quite long, so I'd like to put some more tests into the extra 'longrunning' stage.


spoenemann:
Time fraction of the longrunning tests stage:
 * &asymp; 19% before moving
 * &asymp; 25% after moving

PR #128, opened 2017-03-29
[xbase] add method to allow obtaining return types only.

svenefftinge:


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/IResolvedTypes.java
```diff
@@ -67,12 +67,23 @@
...
+	 * 
 	 * @param expression expression whose return type is queried.
 	 */
 	/* @Nullable */
 	LightweightTypeReference getReturnType(XExpression expression);
 	
 	/**
+	 * Returns the actually inferred return type for the given expression or <code>null</code> if none.
+	 * @param expression expression whose return type is queried.
+	 */
```
cdietrich: do we need a `@since 2.12`

svenefftinge: no, as it is not api

PR #140, opened 2017-05-05
[409297][hightlighting][deprecation] Semantics different from JDTs

cdietrich:
[409297][hightlighting][deprecation] Semantics different from JDTs

### DIFF
--- a/org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/DeprecationUtil.java
+++ b/org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/DeprecationUtil.java
@@ -21,9 +21,20 @@ import org.eclipse.xtext.common.types.JvmType;
  * @author Sebastian Zarnekow - Use explicit deprecated bit if available
  */
 public class DeprecationUtil {
+	
 	public static boolean isDeprecated(JvmAnnotationTarget jvmAnnotationTarget) {
 		if (jvmAnnotationTarget instanceof JvmMember) {
 			return isDeprecatedMember((JvmMember) jvmAnnotationTarget);
+		}
+		return false;
+	}
+	
+	/**
+	 * @since 2.13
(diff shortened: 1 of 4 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=409297

cdietrich:
please review. see https://github.com/eclipse/xtext-xtend/pull/209 as well

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/DeprecationUtil.java
```diff
@@ -23,7 +23,8 @@
 public class DeprecationUtil {
 	public static boolean isDeprecated(JvmAnnotationTarget jvmAnnotationTarget) {
 		if (jvmAnnotationTarget instanceof JvmMember) {
-			return isDeprecatedMember((JvmMember) jvmAnnotationTarget);
+			return isDeprecatedMember((JvmMember) jvmAnnotationTarget)
```
szarnekow: The semantics is now the same as for #isTransitivelyDeprecated. Why doesn't the caller use that one instead?

cdietrich: ill have a look

cdietrich:
i updated the pullrequest @szarnekow

szarnekow:
I wonder if we should make it symmetric and more explicit by adding isTransitivelyDeprecated(JvmAnnotationTarget) and calling that from the highlighting calculator?

cdietrich:
good question. can we make this somehow backwards compatible api wise?

szarnekow:
Adding a new method is not backwards compatible. Apart from that, it would keep the behavior consistent with the prev version.

PR #141, opened 2017-05-12
[516321]  Xtend 2.11 compiles 'forever'

cdietrich:
[516321]  Xtend 2.11 compiles 'forever'

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
@@ -659,7 +659,7 @@ public abstract class AbstractXbaseCompiler {
 	 * @deprecated use {@link #isVariableDeclarationRequired(XExpression, ITreeAppendable, boolean)} instead.
 	 */
 	@Deprecated
-	protected boolean isVariableDeclarationRequired(XExpression expr, ITreeAppendable b) {
+	protected final boolean isVariableDeclarationRequired(XExpression expr, ITreeAppendable b) {
 		return isVariableDeclarationRequired(expr, b, true);
 	}
 
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
@@ -656,11 +656,24 @@ public abstract class AbstractXbaseCompiler {
 	 * If an expression has side effects this method must return true for it.
 	 * @param expr the checked expression
 	 * @param b the appendable which represents the current compiler state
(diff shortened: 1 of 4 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=516321

cdietrich:
please review (for tests: https://github.com/eclipse/xtext-xtend/pull/214)

svenefftinge:
I left a comment on the bugzilla. Explanation for the argument would be good. Otherwise we should go with your change for now.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
```diff
@@ -656,8 +656,9 @@ protected LightweightTypeReference getTypeForVariableDeclaration(XExpression exp
 	 * If an expression has side effects this method must return true for it.
 	 * @param expr the checked expression
 	 * @param b the appendable which represents the current compiler state
+	 * @param recursive TODO
```
cdietrich: Eclipse add param refactoring ftw ;)

svenefftinge:
I think we should overload and deprecate the function with the old signature (delegating with true), because there are xbase clients who might have overridden the method.

cdietrich:
well i would have to delegate to the old method.
and that is recursive ....

cdietrich:
or do you mean calling clients?

svenefftinge:
I mean delegate to the old recursive method with `true`, which should result in the old behaviour.

svenefftinge:
Ah sorry you are right it wouldn't be called anyway

cdietrich:
@svenefftinge pill request updated

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
```diff
@@ -656,11 +656,24 @@ protected LightweightTypeReference getTypeForVariableDeclaration(XExpression exp
 	 * If an expression has side effects this method must return true for it.
 	 * @param expr the checked expression
 	 * @param b the appendable which represents the current compiler state
+	 * @deprecated use {@link #isVariableDeclarationRequired(XExpression, ITreeAppendable, boolean)} instead.
 	 */
+	@Deprecated
 	protected boolean isVariableDeclarationRequired(XExpression expr, ITreeAppendable b) {
```
svenefftinge: If we make this method final, we will force overriding clients to take action, when recompiling their source code.
Which is important, because their overridden implementation will not be called anymore.

I checked github and there are around six overrides of isVariableDeclarationRequired that not only delegate to super. None of them seem to be on the Eclipse train, so they should be able to fix it.

cdietrich: done

Issue #145, opened 2017-06-01
name may not be null xtext 2.12


capacman:
2.12 and throws below exception when doing content assist. HDslParsingTest under henkan dsl project can compile successfully. I see this exception when i run my project with RunWebSocketServer and use vscode. Project and associated forum topic is below

https://www.eclipse.org/forums/index.php/t/1086556/
[henkan.dsl.parent.tar.gz](https://github.com/eclipse/xtext-extras/files/1045029/henkan.dsl.parent.tar.gz)

Caused by: java.lang.UnsupportedOperationException: The resource 'file:///home/capacman/programlama/xtext-workspace/henkan.dsl.parent/vscode-hdsl-ext/exampleproject/example.hl' has been loaded without tracking the feature scope. Please check the implementation of IFeatureScopeTracker.Provider.
at org.eclipse.xtext.xbase.typesystem.internal.IFeatureScopeTracker$1.getExpressionScope(IFeatureScopeTracker.java:63)
at org.eclipse.xtext.xbase.typesystem.internal.ResolvedTypes.getExpressionScope(ResolvedTypes.java:1507)
at org.eclipse.xtext.xbase.typesystem.internal.CompoundReentrantTypeResolver.getExpressionScope(CompoundReentrantTypeResolver.java:180)
at org.eclipse.xtext.xbase.typesystem.internal.ForwardingResolvedTypes.getExpressionScope(ForwardingResolvedTypes.java:134)
at org.eclipse.xtext.xbase.ide.contentassist.XbaseIdeContentProposalProvider.createReceiverProposals(XbaseIdeContentProposalProvider.java:606)
at org.eclipse.xtext.xbase.ide.contentassist.XbaseIdeContentProposalProvider.completeBinaryOperation(XbaseIdeContentProposalProvider.java:535)
at org.eclipse.xtext.xbase.ide.contentassist.XbaseIdeContentProposalProvider._createProposals(XbaseIdeContentProposalProvider.java:377)
at org.eclipse.xtext.xbase.ide.contentassist.XbaseIdeContentProposalProvider.createProposals(XbaseIdeContentProposalProvider.java:641)
at org.eclipse.xtext.ide.editor.contentassist.IdeContentProposalProvider.createProposals(IdeContentProposalProvider.java:94)
at org.eclipse.xtext.ide.server.contentassist.ContentAssistService.createProposals(ContentAssistService.java:108)
[...]

PR #146, opened 2017-06-02
Do not return cached proxies from typeByQueryString-cache. #145

cdietrich:
The elements in the cache might become proxies e.g. if the elements
are unloaded by JvmElementsProxifyingUnloader on
JvmModelAssociator.discardDerivedState(DerivedStateAwareResource)

Returning proxies then will lead to follow up errors as described in #145


cdietrich:
@szarnekow please review

szarnekow:
👍

PR #147, opened 2017-06-02
Added missing @Override

szarnekow:
Added Override annotation to Xbase/QualifiedName to avoid compiler warning.


szarnekow:
@spoenemann Was there a reason why the override annotation was not added before?

cdietrich:
i simply think "forgot"

PR #151, opened 2017-06-13
Cleanup to get rid of a few compiler warnings.

szarnekow:


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/XbaseRuntimeModule.xtend
```diff
@@ -14,7 +14,7 @@ package org.eclipse.xtext.xbase
  */
 class XbaseRuntimeModule extends AbstractXbaseRuntimeModule {
 
-	@SuppressWarnings("deprecation") override Class<? extends org.eclipse.xtext.formatting.IFormatter> bindIFormatter() {
```
cdietrich: should we have a look if we can delete that methods?

szarnekow: Yes, that would be good. But I'd like to investigate this in a separate PR.

szarnekow:
Is this ready to be merged? Any objections?

kthoms:
LGTM

PR #158, opened 2017-07-05
Don't do indentity edits during organize imports.

cdietrich:
Don't do indentity edits during organize imports.
fixes https://github.com/eclipse/xtext-xtend/issues/130

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/imports/ImportOrganizer.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/imports/ImportOrganizer.java
@@ -10,6 +10,7 @@ package org.eclipse.xtext.xbase.imports;
 import static com.google.common.collect.Lists.*;
 import static org.eclipse.xtext.util.Strings.*;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

cdietrich:
see https://github.com/eclipse/xtext-eclipse/pull/296/files as well

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/imports/ImportOrganizer.java
```diff
@@ -56,7 +57,16 @@
 		if (unresolvedTypeResolver != null)
 			unresolvedTypeResolver.resolve(typeUsages, resource);
 		Map<String, JvmDeclaredType> name2type = conflictResolver.resolveConflicts(typeUsages, nonOverridableTypesProvider, resource);
-		return getOrganizedImportChanges(resource, name2type, typeUsages);
+		List<ReplaceRegion> changes = getOrganizedImportChanges(resource, name2type, typeUsages);
```
szarnekow: This looks fishy to me. Modifying the returned list looks not right and a method called 'getXYZChanges' should probably not return things that are no changes at all? Would it be too cumbersome to avoid these replaceRegions in the returned list in the first place?

cdietrich: unfortunatrly the whole code is so bad that you would have throw it completely away and redo it.

szarnekow: Maybe simply move the filtering to the called method?

szarnekow: But it's private anyways. I see your point.

cdietrich: or should i factor out the filtering?

szarnekow: I think it's ok the way it is. A method 'removeNullEdits' would be a plus, though.

PR #160, opened 2017-07-12
[494535]  [xtend][validation] Wrong constant condition check with primitive types

cdietrich:
[494535]  [xtend][validation] Wrong constant condition check with primitive types

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/controlflow/ConstantConditionsInterpreter.xtend
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/controlflow/ConstantConditionsInterpreter.xtend
@@ -355,6 +355,17 @@ class ConstantConditionsInterpreter {
 	}
 	
 	def dispatch EvaluationResult internalEvaluate(XCastedExpression expression, EvaluationContext context) {
+		val declaredType = context.resolvedTypes.getActualType(expression.target)
+		if (declaredType?.primitive) {
+			val castedType = context.resolvedTypes.getActualType(expression)
+			if (castedType !== null && castedType.isPrimitive) {
+				val targetResult = expression.target.doEvaluate(context)
+				if (targetResult.rawValue instanceof Number) {
+					val castedValue = constantOperators.cast(targetResult.rawValue as Number, castedType.identifier)
+					return new EvaluationResult(castedValue,true)
+				}
+			}
+		}
 		return doEvaluate(expression.target, context)
 	}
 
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=494535

cdietrich:
@szarnekow do you know a cleaner way to do this?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/controlflow/ConstantConditionsInterpreter.xtend
```diff
@@ -355,6 +355,13 @@ class ConstantConditionsInterpreter {
 	}
 	
 	def dispatch EvaluationResult internalEvaluate(XCastedExpression expression, EvaluationContext context) {
+		val declaredType = context.resolvedTypes.getActualType(expression.target)
+		if (declaredType?.primitive) {
+			val castedType = context.resolvedTypes.getActualType(expression)
+			if (castedType?.identifier != declaredType?.identifier) {
```
szarnekow: DeclaredType is definitly not null here (see null-check above).
It's safe to use castedType?.type != declaredType.type.

My problem with this solution is, that JDT reports `if ((byte) 1 == (short) 1) ..` as a constant condition. Do we need to check value ranges and floating / integral number casts, instead?

cdietrich: hmmmm good question

cdietrich: you mean something like
```
(byte) 257 == (short) 1
```

szarnekow: Yes, smth along these lines. Or `int i = 20_000; short s = 20_000; (short) i == s;`

cdietrich: have to dive deeper in the code first

cdietrich: hmmm jdt seems not to be so sophisticated

```
byte b = (byte) 257;
		short s = (short) 1;
		if (b == s) {
			
		}
```

cdietrich:
unfortunately == does not handle the implicit cast

=> byte 1 != short 1 but in java is, thus the message that constant expression is false is actually false. dont have an idea how to solve this right now

Issue #164, opened 2017-08-02
SerializerScopeProvider.getExecutableScope prefers getter syntax over property syntax


cdietrich:
SerializerScopeProvider.getExecutableScope prefers getter syntax over property syntax

is this intended?

szarnekow:
I think is is by mistake.

cdietrich:
fixed in 2.13

PR #165, opened 2017-08-03
prefer property syntax over getter syntax

cdietrich:
prefer property syntax over getter syntax
https://github.com/eclipse/xtext-extras/issues/164


cdietrich:
please review

PR #171, opened 2017-08-30
[473623] Xbase creates Imports when refactoring even if the language has no XImport Section Support

cdietrich:
[473623] Xbase creates Imports when refactoring even if the language has no XImport Section Support


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473623

cdietrich:
please review. generator part is here: https://github.com/eclipse/xtext-core/pull/448

cdietrich:
https://github.com/eclipse/xtext-eclipse/pull/340 fix of the refactoring.
the alternative would be to do this more local to the referenceupdater
@JanKoehnlein  @szarnekow what do you think

PR #172, opened 2017-09-04
[521819] Added null guard for field currentThisType

kthoms:
Field might not been initialized when evaluating the condition


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=521819

PR #177, opened 2017-09-07
fixed declaration of inline annotation in xtend #256

cdietrich:


cdietrich:
test at https://github.com/eclipse/xtext-xtend/pull/257

PR #187, opened 2017-10-09
[525746] Set lower bound for org.eclipse.xtext.xbase.lib

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=525746

org.eclipse.xtext.xbase/META-INF/MANIFEST.MF
```diff
@@ -11,7 +11,7 @@ Require-Bundle: org.eclipse.xtext,
  org.eclipse.emf.ecore;bundle-version="2.10.2",
  org.eclipse.emf.common;bundle-version="2.10.1",
  org.antlr.runtime;bundle-version="[3.2.0,3.2.1)",
- org.eclipse.xtext.xbase.lib,
+ org.eclipse.xtext.xbase.lib;bundle-version="2.11.0",
```
cdietrich: why 2.11?

kthoms: No API was added afterwards, so lower bound is OK

kthoms: But actually, for xtext-lib 2.13 makes more sense

cdietrich: API change in 213 were breaking

kthoms: Which change was breaking so that o.e.x.xbase would not work with? The breaking change was StringConcatination#append(String) which was introduced with 2.11. If I see it correctly, xbase does not use API introduced after 2.11. What am I missing?

cdietrich: There were the errors we had recently https://www.eclipse.org/forums/index.php/t/1088834/

kthoms: This does not give me a clue that actually 2.13 is required lower bound. Correct, tooling must run with newer xbase.lib. But 2.11 should be fine, although always recommended to use the same version. The bound defines the lower constraint, not the recommended.

cdietrich: Have no idea and can’t test right now

cdietrich: But won’t you get a method not found exception if the new overloaded methods are not resent

kthoms: Sorry, you are right. There are new API methods without @since tag. Thus I did not see that. I will create a separate PR for this and then update constraints accordingly.

kthoms: Problematic were especially the new no-args methods in CollectionLiterals, which were used all over the code base.

kthoms: Updated to lower 2.13, also for other PRs

kthoms: Thanks for being insistent on this!

PR #193, opened 2017-10-24
Ignore default methods from superclass branch when checking ambiguous inherited methods.

gallandarakhneorg:
This patch fixes an incomplete coverage of JLS 8.4.8
(https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8).
Overriding conflicts were detected between interfaces from the direct
superclass branch and a direct superinterface. It is not compliant with
the JLS and the current behavior of javac.

Complete unit tests are is [Xtend PR 192](https://github.com/eclipse/xtext-xtend/pull/192).

see https://github.com/eclipse/xtext/issues/2296
see https://github.com/eclipse/xtext-xtend/pull/192

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/override/ResolvedFeatures.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/override/ResolvedFeatures.java
@@ -10,6 +10,7 @@ package org.eclipse.xtext.xbase.typesystem.override;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
@@ -39,6 +40,7 @@ import com.google.common.collect.Sets;
 /**
  * @author Sebastian Zarnekow - Initial contribution and API
  * @author Lorenzo Bettini - https://bugs.eclipse.org/bugs/show_bug.cgi?id=454786
+ * @author Stephane Galland - https://github.com/eclipse/xtext-xtend/issues/191
  */
 public class ResolvedFeatures extends AbstractResolvedFeatures {
 
@@ -250,6 +252,9 @@ public class ResolvedFeatures extends AbstractResolvedFeatures {
 
 			private Set<JvmType> seen = Sets.newHashSet();
 			
+			private boolean isSuperclassBranch;
+			private Multimap<String, AbstractResolvedOperation> superclassBranchOperations;
+			
 			@Override
 			public Boolean doSwitch(EObject theEObject) {
 				if (theEObject == null)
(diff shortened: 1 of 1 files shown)


szarnekow:
Did you try to play with `org.eclipse.xtext.xbase.typesystem.override.ResolvedFeatures.handleOverridesAndConflicts(JvmOperation, Multimap<String, AbstractResolvedOperation>)`, e.g. line 361 ( https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/override/ResolvedFeatures.java#L361 ) where we could pass true instead of false? Would that make a difference for your usecase?

gallandarakhneorg:
Changing within the original code only the value of the argument from `false` to `true` does not solve the issue. Output error is: `The type A inherits multiple implementations of the method spliterator() from Iterable and Set.` The conflicting `Iterable` comes from the `AbstractCollection` inheritance branch.

Reminder of the tested code:
```xtend
import java.util.Iterator
import java.util.Set
import java.util.AbstractCollection
class A<T> extends AbstractCollection<T> implements Set<T> {
  override Iterator<T> iterator {
    null
  }
  override int size {
    0
  }
}
```
According to my interpretation of JLS: `Set` should provides the default implementation for `splititerator` because it a direct super type of that provides this default implementation. And, in fact, javac does not fail on this code.

szarnekow:
Understood. So what is missing in the current impl is the concept of a direct supertype. Are there other references to it in the JLS?

gallandarakhneorg:
I think it is more related to the difference between the superclass inheritance and superinterface inheritance than the concept of "direct" super type. I have written the patch according to this point of view. I think the key sentence is "[It is a compile-time error if a class C inherits a default method whose signature is override-equivalent with another method inherited by C, unless there exists an abstract method declared in a superclass of C and inherited by C that is override-equivalent with the two methods.](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.4)" I have not found a clear statement within thr JLS regarding the observed behavior of javac, except in this section and Section 8.4.8. If you find better explanation, please give me back.

It may be interesting to write a systematic test bench of all possible cases, and compare the answers of javac and Xtend.

szarnekow:
> It may be interesting to write a systematic test bench of all possible cases, and compare the answers of javac and Xtend.

Would you have the time to do that?

gallandarakhneorg:
Under process: I have a first script generating more than 8000 unit tests.

szarnekow:
Hey @gallandarakhneorg, did you make more progress on this ticket?

gallandarakhneorg:
Yes. But, it is still not finish (validated of the generated test coverage). I may provide inputs for next week.

gallandarakhneorg:
I have a script that creates 8747 unit tests related to different inheritance cases.
Because I'm using Xtend for creating the tests, I think this project is not the best place to put them.
Should I request a pull in Xtend project? Or is another way to use Xtend for tests within xtext-extras?

szarnekow:
That sounds great. Yes, you are right, these tests should be put into the Xtend unit test suites, if they use the Xtend language to test the inheritance / override validation.

gallandarakhneorg:
Ok, I'm validating my generated unit tests on my own DSL, which includes Xtend, in order to detect the failures and fix the minor things. Then, I will request a pull to Xtend project and give the reference within this discussion.

szarnekow:
That's a sound plan. 👍 Very good! Thank you for the effort!

gallandarakhneorg:
Unit tests are discussed within https://github.com/eclipse/xtext-xtend/pull/192

cdietrich:
i assume this branch and the test branch both need a rebase

szarnekow:
I'm not sure how to proceed with the effort that was spent on this issue.

cdietrich:
move to 2.24

cdietrich:
this repository has been merged/moved to https://github.com/eclipse/xtext.
please feel free to recreate this PR against that repository in case it still is relevant.

Issue #199, opened 2017-11-28
Make FeatureScopeTracker instantiable without creating an extra subclass


cdietrich:

kthoms:
I assume you have a concrete use case?

cdietrich:
https://github.com/eclipse/xtext/issues/2359
https://github.com/sjka/smarthome/commit/aa35e658ad146ca445798ce860d11da111d9028e
plus something similar in customer project

cdietrich:
Fixed in 2.14

PR #200, opened 2017-11-28
made constructor of FeatureScopeTracker public #199

cdietrich:
made constructor of FeatureScopeTracker public so it can be instantiated more easily by subclasses of OptimizingFeatureScopeTrackerProvider #199


cdietrich:
please review

PR #205, opened 2017-12-06
Regenerated Xbase

szarnekow:


org.eclipse.xtext.xbase.tests/longrunning/src/org/eclipse/xtext/xbase/tests/typesystem/TimedBatchTypeResolverTest.xtend
```diff
@@ -60,7 +61,7 @@ class TimedBatchTypeResolverTest extends AbstractBatchTypeResolverTest {
 class TypeResolverPerformanceTest extends BatchTypeResolverTest {
 	
 	@Rule
-	public val timeout = new Timeout(100) // TODO improve - aim at something like 100
+	public val timeout = new Timeout(100, TimeUnit.MILLISECONDS) // TODO improve - aim at something like 100
```
cdietrich: comment makes no sense

org.eclipse.xtext.xbase/generator/org/eclipse/xtext/xbase/GenerateXbase.java
```diff
@@ -39,7 +39,7 @@ public static void main(String[] args) {
 		final String root = "..";
 		final String projectName = "org.eclipse.xtext.xbase";
 		final String runtimeProject = root + "/" + projectName;
-		final String uiProject = root + "/../xtext-eclipse/" + projectName + ".ui";	
+		final String uiProject = System.getProperty("uiProject", root + "/../xtext-eclipse/" + projectName + ".ui");
```
cdietrich: hmmm i wonder why this is needed. shouldnt root be the parameter as it is for other workflows?

szarnekow: runtimeProject and uiProject are in different repositories and there is no reliable way to navigate in the file system of the developers machine from one repo to another. Thus I need to pass the other projects path as an argument

org.eclipse.xtext.xbase.tests/longrunning/src/org/eclipse/xtext/xbase/tests/typesystem/TimedBatchTypeResolverTest.xtend
szarnekow: The test class is ignored because the timout is too strict. But yes, the comment is not too helpful.

PR #243, opened 2018-03-24
[460455] Avoid fully qualifying names when organizing imports.

miklossy:
- Modify the Xbase ConflictResolver not to fully qualify type names if
two types (one as inner type, one as top-level type) with the same name
exists. In such case, the usage of the fully qualified name is
unnecessary, since the inner type has precedence anyway.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=460455

PR #257, opened 2018-06-03
[eclipse/xtext-xtend#461] Avoid stackoverflow in type inferrence

szarnekow:
We didn't handle more involving generics situtations properly, e.g. when type constraints referred to type parameters of another nesting level, e.g `T extends A & B, X extends I<T>` where `I<T>`refers to a generic type with complex type constraints, too. During inference, we do now detect situations where we again compare potential type substitutes even though we had seen those before.
The fix is twofold: We figure during the processing of UnboundTypeReferences in the RawTypeConformanceComputer that we did already process a certain reference. The second piece to the puzzle is that we do not collect substitution hints when we know for sure that they are invalid, e.g. when they do not match the type constraints.

This PR also contains minor cosmetic fixes, e.g. for error messages or the instantiation of TypeConformanceComputationArguments. Tests are in the Xtend repo. PR follows.

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/RawTypeConformanceComputer.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/RawTypeConformanceComputer.java
@@ -632,13 +632,8 @@ public class RawTypeConformanceComputer {
 		if (leftHints.isEmpty() && (flags & UNBOUND_COMPUTATION_ADDS_HINTS) == 0) {
 			return flags; 
 		}
-		for(LightweightBoundTypeArgument leftHint: leftHints) {
-			if (leftHint.getSource() == BoundTypeArgumentSource.INFERRED_LATER && leftHint.getActualVariance() == VarianceInfo.INVARIANT) {
-				LightweightTypeReference leftHintReference = leftHint.getTypeReference();
-				if (leftHintReference.getUniqueIdentifier().equals(right.getUniqueIdentifier())) {
-					return flags | SUCCESS;
-				}
-			}
+		if (isConstrainedRecursiveHintCheck(leftHints, right)) {
+			return flags | SUCCESS;
 		}
 		int result = isConformantToConstraints(left, right, leftHints, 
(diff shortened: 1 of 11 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/RawTypeConformanceComputer.java
```diff
@@ -649,6 +652,25 @@ LightweightMergedBoundTypeArgument getMergeResult(UnboundTypeReference left) {
 		return flags;
 	}
 	
+	private boolean isConstrainedRecursiveHintCheck(List<LightweightBoundTypeArgument> leftHints, LightweightTypeReference right) {
+		boolean hasConstraints = false;
+		for(LightweightBoundTypeArgument leftHint: leftHints) {
```
cdietrich: how are these left hints sorted?

szarnekow: They are returned in the order of their appearance. The first hints are usually deduced from the type parameter declarations.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/TypeConformanceComputationArgument.java
```diff
@@ -13,6 +13,9 @@
  */
 public class TypeConformanceComputationArgument {
 
+	public static final TypeConformanceComputationArgument RAW = new TypeConformanceComputationArgument(true, false, true, true, false, true);
+	public static final TypeConformanceComputationArgument DEFAULT = new TypeConformanceComputationArgument();
+
```
cdietrich: should the constructor with no arguments be private?

szarnekow: I considered that, too but was a little afraid that clients may use it already in their code.

cdietrich: good point. maybe mark it as deprecated?

szarnekow: :+1: Done

PR #258, opened 2018-06-04
[eclipse/xtext-xtend#447] dont issue warnings for raw types inside instanceof

cdietrich:
[eclipse/xtext-xtend#447] dont issue warnings for raw types inside instanceof


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/JvmTypeReferencesValidator.java
```diff
@@ -129,6 +131,9 @@ public void checkTypeArgsAgainstTypeParameters(JvmParameterizedTypeReference typ
 	}
 
 	protected void warnRawType(JvmType type, JvmParameterizedTypeReference typeRef) {
+		if (EcoreUtil2.getContainerOfType(typeRef, XInstanceOfExpression.class) != null) {
```
szarnekow: We should check only the immediate container. Otherwise we will also miss warnings in code like 
```
{ var Map x = null x } instanceof Map<?, ?>
```
or 
```
a(x as Map) instanceof List<?>
```

cdietrich:
@szarnekow updated the pr

PR #260, opened 2018-06-05
Reduced "potential null pointer access" errors

tivervac:


tivervac:
This is of course valid and shorter xtend code but the change doesn't reduce readability too much and gets rid of the jdt error

PR #263, opened 2018-06-15
[#262] Made type lookup case sensitive even on windows

szarnekow:
This appears to be a rather expensive solution for a rare corner case but I'm not sure how to solve this in another way. The file / resource lookup is completely transparent to Java and handled natively by the OS. That's why we need to convert the found name to the canonical name to check if it's the real thing.

### DIFF
--- a/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/BinaryClassFinderTest.java
+++ b/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/BinaryClassFinderTest.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.common.types.access.impl;
 
+import org.eclipse.xtext.common.types.access.binary.BinaryClass;
 import org.eclipse.xtext.common.types.access.binary.BinaryClassFinder;
 import org.junit.After;
 import org.junit.Assert;
@@ -56,6 +57,27 @@ public class BinaryClassFinderTest extends Assert {
 	
 	@Test public void testObjectArray_02() throws ClassNotFoundException {
 		assertEquals(Object[].class.getName(), classFinder.forName(Object[].class.getName()).getName());
+	}
+	
+	@Test public void testBinaryClassFinder() throws ClassNotFoundException {
+		assertEquals(BinaryClassFinder.class.getName(), classFinder.forName(BinaryClassFinder.class.getName()).getName());
+	}
+	
+	@Test(expected=ClassNotFoundException.class) public void testBinaryClassFinderWrongCasing() throws ClassNotFoundException {
+		classFinder.forName(BinaryClassFinder.class.getName().toUpperCase());
+	}
+	
+	@Test public void testBinaryClassFinderWrongCasingForcedCaseSensitive() throws ClassNotFoundException, NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException {
+		BinaryClass.assumeCaseSensitive((originalValue)->{
+			try {
+				String upperCase = BinaryClassFinder.class.getName().toUpperCase();
+				BinaryClass found = classFinder.forName(upperCase);
+				assertFalse(originalValue);
+				assertEquals(upperCase, found.getName());
+			} catch(ClassNotFoundException e) {
+				assertTrue(originalValue);
+			}
+		});
 	}
 	
 }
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/access/binary/BinaryClass.java
```diff
@@ -228,6 +230,35 @@ public Throwable fillInStackTrace() {
...
+			File tempFile = File.createTempFile("prefix", "");
+			File upperCase = new File(tempFile.getParentFile(), tempFile.getName().toUpperCase());
+			return !(upperCase.exists() && upperCase.getCanonicalFile().getName().equals(tempFile.getName()));
+		} catch(IOException ex) {
+			return false;
+		}
+	}
+	
+	private static boolean fileSystemAwareExists(URL url) {
+		if (!IS_CASE_SENSITIVE && "file".equals(url.getProtocol())) {
```
cdietrich: do we have estimates how the penalty for this is

szarnekow: Unfortunately not. I’m afraid it’s expensive though. That’s why I introduced a means to skip the check.

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/access/binary/BinaryClass.java
```diff
@@ -228,6 +230,35 @@ public Throwable fillInStackTrace() {
...
 	}
+	
+	private static final boolean IS_CASE_SENSITIVE = isCaseSensitive();
+	
+	private static boolean isCaseSensitive() {
+		String result = System.getProperty("org.eclipse.xtext.common.types.access.binary.BinaryClass.ASSUME_CASE_SENSITIVE", Boolean.toString(inferIsCaseSensitive()));
+		return Boolean.parseBoolean(result);
+	}
+	
+	private static boolean inferIsCaseSensitive() {
```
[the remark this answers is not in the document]
szarnekow: I did not find a helper in any of our current 3rd party libs. I did not want to introduce a new dependency thus this approach.

szarnekow:
I added some tests.

cdietrich:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.eclipse.xtext.common.types.access.impl.BinaryClassFinderTest.lambda$testBinaryClassFinderWrongCasingForcedCaseSensitive$0(BinaryClassFinderTest.java:78)
	at org.eclipse.xtext.common.types.access.binary.BinaryClass.assumeCaseSensitive(BinaryClass.java:244)
	at org.eclipse.xtext.common.types.access.impl.BinaryClassFinderTest.testBinaryClassFinderWrongCasingForcedCaseSensitive(BinaryClassFinderTest.java:71)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflec
```

cdietrich:
(mac and windows are case sensitive => am not sure if it makes since to test for false)

cdietrich:
in the tests the classes are inside the jar so the tolowercase class is not found.
=> i propose to use `BinaryClassFinderTest` as testclass

szarnekow:
Alternatively we could load the wrong class file and check the name / package afterwards. This would safe us from the additional lookup for the common case. I'll provide an updated patch.

org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/BinaryClassFinderTest.java
```diff
@@ -60,17 +60,17 @@ public void tearDown() throws Exception {
 	}
 	
 	@Test public void testBinaryClassFinder() throws ClassNotFoundException {
-		assertEquals(BinaryClassFinder.class.getName(), classFinder.forName(BinaryClassFinder.class.getName()).getName());
+		assertEquals(BinaryClassFinderTest.class.getName(), classFinder.forName(BinaryClassFinderTest.class.getName()).getName());
```
cdietrich: :+1:

szarnekow:
This will unlikely go into 2.16 I suppose.

cdietrich:
@szarnekow can we close this pr and leave it on the branch

Issue #261, opened 2018-06-12
JavaSource Language is very inperformant


cdietrich:
the JavaSource Language is very inperformant e.g. when used in the context of the xtend gradle plugin (https://github.com/xtext/xtext-gradle-plugin/issues/89).

The main performance downside in the `installFull` in  `JavaDerivedStateComputer` which makes inefficient use of a own `IndexAwareNameEnvironment` and the jdt compiler

here is a small but "unfair" sample that shows the problem quite nicely i think

```

import com.google.inject.Inject
import com.google.inject.Provider
import java.util.List
import org.eclipse.emf.common.util.URI
import org.eclipse.xtext.common.types.access.IJvmTypeProvider
import org.eclipse.xtext.java.JavaSourceLanguageSetup
import org.eclipse.xtext.resource.IResourceDescription
import org.eclipse.xtext.resource.XtextResourceSet
import org.eclipse.xtext.resource.impl.ChunkedResourceDescriptions
import org.eclipse.xtext.resource.impl.ResourceDescriptionsData
import org.eclipse.xtext.testing.util.InMemoryURIHandler

class DummyTest {
	
	def static  void main(String[] args) {
		val injector = new JavaSourceLanguageSetup().createInjectorAndDoEMFRegistration
		val dt = injector.getInstance(DummyTest)
		dt.doit()
	}
	
	    @Inject Provider<XtextResourceSet> resourceSetProvider
    @Inject IResourceDescription.Manager resourceDesriptionManager
    @Inject IJvmTypeProvider.Factory typeProviderFactory
	
	def void doit() {
		val max = 10000
		val rs = (1..max).map["demo/Demo"+it+".java"->'''package demo; public class Demo«it» {
			private String x;
			«IF it > 1»
			private Demo«it-1» demo;
			«ENDIF»
			«IF it > 2»
			private Demo«it-2» demo;
			«ENDIF»
			«IF it < max-1»
			private Demo«it+1» demo;
			«ENDIF»
		}'''].toList.resourceSet
		println("creation done")
		for (r : rs.resources) {
			r.contents
		}
	}
	
	    def protected resourceSet(Pair<String, String> ... files) {
        val result = resourceSetProvider.get
        typeProviderFactory.createTypeProvider(result)
        result.classpathURIContext = class.classLoader
        result.URIConverter.URIHandlers.clear
        val uriHandler = new InMemoryURIHandler
        val uris = files.map [
[...]

PR #264, opened 2018-06-18
[#261] Implemented a Binary Class Lookup Cache for Java Sources

cdietrich:
[#261] Implemented a Binary Class Lookup Cache for Java Sources


cdietrich:
this is just a pull request to discuss possible solutions for #261 
@szarnekow would be nice if we could have a look somewhen next week

cdietrich:
still the question: how can this be performance and function tested locally.

szarnekow:
My mac file system is case sensitive when it matters and case insensitive in general. So functional testing was already done. Regarding perf tests - I'm still thinking about a different approach that would avoid the double check on all class files. But that is not yet green.

cdietrich:
TODO: cache the ibinarytype and not the answers

PR #279, opened 2018-07-23
fixed broken javadoc

cdietrich:

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/IFeatureScopeTracker.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/IFeatureScopeTracker.java
@@ -37,12 +37,12 @@ public interface IFeatureScopeTracker {
 	}
 	
 	/**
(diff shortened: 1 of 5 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/IFeatureScopeTracker.java
```diff
@@ -37,12 +37,12 @@
 	}
 	
 	/**
-	 * @see IResolvedTypes#getExpressionScope(EObject, Anchor)
+	 * @see IResolvedTypes#getExpressionScope(EObject, IExpressionScope.Anchor)
```
kthoms: wondering why this qualification is necessary, since Anchor is imported. But if you saw an error before...

cdietrich: thats why i have changed that

szarnekow: Nested types cannot be referenced directly from JavaDoc as far as I know.

PR #283, opened 2018-07-25
Fixed extension name computation of getters and is-methods in the SerializerScopeProvider

tivervac:
The `SerializerScopeProvider` always calculates the name for extension methods by throwing away the `get` or `is` at the start of the function and then making the first remaining character lowercase. This runs into trouble for methods such as `getURI`. That method's extension name is `URI` not `uRI`.

### DIFF
--- a/org.eclipse.xtext.xbase.testing/testdata/testdata/ShorthandNames.java
+++ b/org.eclipse.xtext.xbase.testing/testdata/testdata/ShorthandNames.java
@@ -1,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package testdata;
+
+/**
+ * @author dietrich - Initial contribution and API
+ */
+public class ShorthandNames {
+
+	private int fake;
+	private int UPPER;
+	private int lower;
+	private int camelCase;
+	private int UPper;
+	private int x;
+
+	private boolean UPPERBOOL;
+	private boolean lowerbool;
+	private boolean camelCaseBool;
+	private boolean UPperBool;
+
+	public int get() {
+		return fake;
+	}
+	public void set(int fake) {
+		this.fake = fake;
+	}
+	public boolean is() {
+		return true;
+	}
+	public int getUPPER() {
+		return UPPER;
+	}
+	public void setUPPER(int uPPER) {
+		UPPER = uPPER;
+	}
+	public int getLower() {
+		return lower;
+	}
+	public void setLower(int lower) {
+		this.lower = lower;
+	}
+	public int getCamelCase() {
+		return camelCase;
+	}
+	public void setCamelCase(int cameCase) {
+		this.camelCase = cameCase;
+	}
+	public int getUPper() {
+		return UPper;
+	}
+	public void setUPper(int UPper) {
+		this.UPper = UPper;
+	}
+	public int getX() {
+		return x;
+	}
+	public void setX(int x) {
+		this.x = x;
+	}
+	public boolean isUPPERBOOL() {
+		return UPPERBOOL;
+	}
+	public void setUPPERBOOL(boolean uPPERBOOL) {
+		UPPERBOOL = uPPERBOOL;
+	}
+	public boolean isLowerbool() {
+		return lowerbool;
+	}
+	public void setLowerbool(boolean lowerbool) {
+		this.lowerbool = lowerbool;
+	}
+	public boolean isCamelCaseBool() {
+		return camelCaseBool;
+	}
+	public void setCamelCaseBool(boolean camelCaseBool) {
+		this.camelCaseBool = camelCaseBool;
+	}
+	public boolean isUPperBool() {
+		return UPperBool;
+	}
+	public void setUPperBool(boolean UPperBool) {
+		this.UPperBool = UPperBool;
+	}
+}
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/scoping/batch/AbstractSessionBasedScope.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/scoping/batch/AbstractSessionBasedScope.java
@@ -110,7 +110,7 @@ public abstract class AbstractSessionBasedScope extends AbstractScope {
 	}
 	
 	/**
-	 * Returns the name as a property name, e.g. a prefix {@code get}, {@code is} or {@code set} 
+	 * Returns the name as a property name, e.g. a prefix {@code get}, {@code is} or {@code set}
 	 * can be used with the result of this method.
 	 * If the given name is invalid, the result is <code>null</code>.
 	 */
@@ -145,7 +145,7 @@ public abstract class AbstractSessionBasedScope extends AbstractScope {
 		// length 0 is invalid
 		return null;
 	}
-	
+
 	/**
 	 * Clients may override to reject certain descriptions from the result. All subtypes of {@link AbstractSessionBasedScope}
 	 * in the framework code will delegate to this method to accumulate descriptions in a list.
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/SerializerScopeProvider.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/SerializerScopeProvider.java
@@ -47,6 +47,7 @@ import org.eclipse.xtext.xbase.scoping.batch.IBatchScopeProvider;
 import org.eclipse.xtext.xbase.scoping.batch.IFeatureNames;
 import org.eclipse.xtext.xbase.scoping.featurecalls.OperatorMapping;
 import org.eclipse.xtext.xbase.typesystem.util.IVisibilityHelper;
+import org.eclipse.xtext.xbase.util.PropertyUtil;
 import org.eclipse.xtext.xtype.XImportDeclaration;
 import org.eclipse.xtext.xtype.XImportSection;
 
@@ -153,7 +154,8 @@ public class SerializerScopeProvider implements IScopeProvider, IFeatureNames {
 	}
 
 	protected IScope getExecutableScope(XAbstractFeatureCall call, JvmIdentifiableElement feature) {
-		QualifiedName name = QualifiedName.create(feature.getSimpleName());
+		final String simpleName = feature.getSimpleName();
+		QualifiedName name = QualifiedName.create(simpleName);
 		if (call.isOperation()) {
 			QualifiedName operator = getOperator(call, name);
 			if (operator == null) {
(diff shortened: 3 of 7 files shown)


tivervac:
Fixes https://github.com/eclipse/xtext-xtend/issues/514

cdietrich:
- please investigate the failing tests
- is still wonder if we can unify this with a fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=478898

tivervac:
Will do on monday!

cdietrich:
maybe you can add a testclass like that to testdata
```
/*******************************************************************************
 * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 *******************************************************************************/
package testdata;

/**
 * @author dietrich - Initial contribution and API
 */
public class PropertyNames {
	
	private int UPPER;
	private int lower;
	private int camelCase;
	
	private boolean UPPERBOOL;
	private boolean lowerbool;
	private boolean camelCaseBool;
	public int getUPPER() {
		return UPPER;
	}
	public void setUPPER(int uPPER) {
		UPPER = uPPER;
	}
	public int getLower() {
		return lower;
	}
	public void setLower(int lower) {
		this.lower = lower;
	}
	public int getCamelCase() {
		return camelCase;
	}
	public void setCamelCase(int cameCase) {
		this.camelCase = cameCase;
	}
	public boolean isUPPERBOOL() {
		return UPPERBOOL;
	}
	public void setUPPERBOOL(boolean uPPERBOOL) {
		UPPERBOOL = uPPERBOOL;
	}
	public boolean isLowerbool() {
		return lowerbool;
	}
	public void setLowerbool(boolean lowerbool) {
		this.lowerbool = lowerbool;
	}
	public boolean isCamelCaseBool() {
		return camelCaseBool;
	}
	public void setCamelCaseBool(boolean camelCaseBool) {
		this.camelCaseBool = camelCaseBool;
	}

}

```
and add testcases like these to the serializer tests or baseclass AbstractXbaseEvaluationTest

one per test method
```
assertEvaluatesTo(true, "(new testdata.PropertyNames() => [ lowerbool = true ]).lowerbool");
		assertEvaluatesTo(1, "(new testdata.PropertyNames() => [ lower = 1 ]).lower");
		assertEvaluatesTo(true, "(new testdata.PropertyNames() => [ camelCaseBool = true ]).camelCaseBool");
		assertEvaluatesTo(1, "(new testdata.PropertyNames() => [ camelCase = 1 ]).camelCase");
		assertEvaluatesTo(true, "(new testdata.PropertyNames() => [ UPPERBOOL = true ]).UPPERBOOL");
		assertEvaluatesTo(1, "(new testdata.PropertyNames() => [ UPPER = 1 ]).UPPER");
```

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/SerializerScopeProvider.java
```diff
@@ -162,21 +164,25 @@ protected IScope getExecutableScope(XAbstractFeatureCall call, JvmIdentifiableEl
 			return new SingletonScope(EObjectDescription.create(operator, feature), IScope.NULLSCOPE);
 		}
 		if (call instanceof XAssignment) {
-			String propertyName = Strings.toFirstLower(feature.getSimpleName().substring(3));
+			String propertyName = Strings.toFirstLower(simpleName.substring(3));
```
szarnekow: Here we should use tryGetAsPropertyName, too.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/scoping/batch/AbstractSessionBasedScope.java
```diff
@@ -108,6 +109,43 @@ protected String toProperty(String methodName, JvmFeature feature, int getterPar
 		return PropertyUtil.getPropertyName(feature, methodName, getterParams, setterParams);
 	}
 	
+	/**
+	 * Returns the name as a property name, e.g. a prefix {@code get}, {@code is} or {@code set}
+	 * can be used with the result of this method.
+	 * If the given name is invalid, the result is <code>null</code>.
+	 */
+	/* @Nullable */
+	protected String tryGetAsPropertyName(String name) {
```
tivervac: @szarnekow I reverted this method as it did not do what I thought it did.
The method checks if a given name is valid to be extended (add set/get/is) to it. e.g. Class.CanonicalName is not valid as it doesn't map to Class.getCanonicalName. I added a new method in PropertyUtil that does actual conversion from full function names to their short hand name.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/util/PropertyUtil.xtend
```diff
@@ -50,21 +51,26 @@ class PropertyUtil {
...
+	 * 
 	 * @since 2.15
 	 */
 	/* @Nullable */
-	def static String tryGetAsPropertyName(String name) {
-		if (name.length() == 1) { // e.g. Point.getX()
-			if (Character.isUpperCase(name.charAt(0))) {
-				// X is not a valid sugar for getX()
-				return null;
+	def static String tryGetShorthandName(String fullName) {
```
tivervac: @szarnekow I changed the name here to avoid confusion with the previous tryGetAsPropertyName. Is this ok?

tivervac:
@cdietrich Is there a way we can wrap PropertyUtils (as described in bugzilla) without breaking API?

cdietrich:
@szarnekow do you have any idea on how to make PropertyUtils injectable?

org.eclipse.xtext.xbase.testing/testdata/testdata/ShorthandNames.java
```diff
@@ -0,0 +1,90 @@
...
+	private int camelCase;
+	private int UPper;
+	private int x;
+
+	private boolean UPPERBOOL;
+	private boolean lowerbool;
+	private boolean camelCaseBool;
+	private boolean UPperBool;
+
+	public int get() {
```
szarnekow: tryGetShorthand does contain logic for methods like `getuPperCase`. Can you please double-check that this is in sync with java.beans.Introspector and add a testcase?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/util/PropertyUtil.xtend
szarnekow: We haven't used shorthand before in that context so I'm not 100% convinced.

tivervac: @szarnekow Then what would be the correct terminology?

tivervac: Having looked at the JavaBeans [spec](http://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/) in the meantime, I believe the correct name would be `inferred property name` (see 8.8 in the spec).

org.eclipse.xtext.xbase.testing/testdata/testdata/ShorthandNames.java
tivervac: I believe `getuPperCase` should allow you to access the getter through `uPperCase`, but this is not the case in Xtend.

Before testing serialization we test that there are no errors in the parse result. There is an error as it thinks you're trying to access a private field instead of a getter method.
As thus, I cannot write the test as the logic is never hit.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/SerializerScopeProvider.java
```diff
@@ -164,26 +164,25 @@ protected IScope getExecutableScope(XAbstractFeatureCall call, JvmIdentifiableEl
 			return new SingletonScope(EObjectDescription.create(operator, feature), IScope.NULLSCOPE);
 		}
 		if (call instanceof XAssignment) {
-			String shorthandName = PropertyUtil.tryGetShorthandName(simpleName);
```
tivervac: @szarnekow Is there a reason that we returned a scope only containing the shortened name here? The scope for setters would always have been `[foo]` instead of `[foo, setFoo]`. For getters, see below, both the exact and the shortened name are added to the scope.

szarnekow: I would say `this.setFoo = 7;` should at least be a warning since this is not what I consider good coding style. Appears to be a bug in the scoping logic. From that POV we should include both variants here.

org.eclipse.xtext.xbase.testing/testdata/testdata/ShorthandNames.java
szarnekow: According to java.beans.PropertyDescriptor.getBaseName() I would think that `getuPper` cannot be aliased by `uPper`

tivervac: Isn't getBase name the other way around though? I think that method is used to generate accessors from variables: `uPper => get/setuPper`

szarnekow: getBaseName is used to compute the basename, e.g. the name after the get prefix. It'll always do an toUpperCase on the first letter. So the read- and write methods cannot be found afterwards since the code will look for getUPper rather than getuPper. So the property access wouldn't work. That is my understanding.

tivervac: So the shortened name for `getuPper` should be `UPper`? That's not how it works in Xtend. In Xtend there is no shortened name for `getuPper` at all. In any case the logic will have to be adjusted

szarnekow: No, actually not. In fact it's not 100% symetrical. The shorthand name for getuPper would be uPper, but later on it would fail to bind the property to the method (if we'd be an interpreted language with reflective bean properties). That's why I'D consider getuPper to be not a valid property accessor.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/util/PropertyUtil.xtend
szarnekow: I like `inferred property name`!

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/SerializerScopeProvider.java
tivervac: @szarnekow This changes behaviour a little bit.
e.g. in [this](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/refactoring/JavaRefactoringIntegrationTest.java#L537) test `setFoo` is refactored to `setBaz` which is supposed to conflict because `foo=null` would become `baz=null` which assigns to the field `baz` instead of calling the new refactored method `setBaz`.
By now adding `setBaz` to the scope we first try refactoring `foo=null` to `baz=null` and fail, then we try refactoring it to `setBaz=null`.

This is valid code but as you said not good coding style. Should we fix this somewhere else, or should we allow refactoring to potentially create new warnings?

szarnekow: I'm ok (for now) if the refactoring produces new warnings. In the future I'd rather like to refactor the assignment to a feature call, e.g. instead of setBaz = null it should become setBaz(null)

cdietrich:
@szarnekow @tivervac what is the state of this pr?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/util/PropertyUtil.xtend
```diff
@@ -48,4 +50,42 @@ class PropertyUtil {
...
+		} else if (name.length() > 1) {
+			if (Character.isUpperCase(name.charAt(1))) { // e.g. Resource.getURI
+				// if second char is uppercase, the name itself is the sugar variant
+				// URI is the property name for getURI
+				if (Character.isUpperCase(name.charAt(0))) {
+					return name;
+				}
+				// if the first character is not upper case, it's not a valid sugar variant
+				// e.g. uRI is no sugar access for getURI
+				return name;
```
szarnekow: The code doesn't match the comment here. Should this return null instead?

tivervac: This comment appears to be copied from the previous utility, the return is correct but I'll fix the comment.

tivervac:
Squashed my changes

PR #284, opened 2018-08-06
[xtext/xtext-gradle-plugin#120] performance optimization java source lang

cdietrich:
[xtext/xtext-gradle-plugin#120] performance optimization java source lang

[#261] Implemented a Binary Class Lookup Cache for Java Sources


org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
```diff
@@ -0,0 +1,44 @@
...
+ * @author Christian Dietrich - Initial contribution and API
+ * @since 2.14
+ */
+@EmfAdaptable
+class ClassFileCache {
+	
+	//TODO: concurrency
+	//TODO: clear
+	//TODO: weak?
+	val Map<QualifiedName, IBinaryType> cache = new HashMap()
```
szarnekow: Should indeed be a concurrentHashMap

szarnekow: My feeling is also that we should use String as the key since it is easier to create the same string for nested types / types with $ in their names compared to equal QN

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -31,7 +32,15 @@ import org.eclipse.emf.ecore.resource.Resource
 		return findType(className)
 	}
 	
-	def findType(QualifiedName className) {
+	def NameEnvironmentAnswer findType(QualifiedName className) {
+		if (classFileCache.containsKey(className)) {
+			val t = classFileCache.get(className)
+			// TODO is this ok?
+			if (t===null) {
```
szarnekow: ConcurrentHashMap cannot contain null. We may need a dedicated null value

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -51,9 +60,13 @@ import org.eclipse.emf.ecore.resource.Resource
 			val url = classLoader.getResource(fileName)
 			if (url === null) {
 				cache.put(className, null)
+				//TODO is that ok
+				classFileCache.put(className, null)
```
szarnekow: See prev comment

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -51,9 +60,13 @@ import org.eclipse.emf.ecore.resource.Resource
...
 				return null;
 			}
 			val reader = ClassFileReader.read(url.openStream, fileName)
+			// TODO is this ok?
+			classFileCache.put(className, reader)
```
szarnekow: Should we store the NameEnvironmentAnswer instead?

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
szarnekow: The cache should than expose methods like putIfAbsent or getOrCreate(key, Supplier)

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
szarnekow: Also see org.eclipse.xtext.common.types.access.impl.CachingDeclaredTypeFactory

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
cdietrich: so you dont like the usage of the active annotation?

szarnekow: No, sorry, that was not what I meant. On the cache itself there is containsKey, put and so on and so forth. That should be a one-shot operation that we can pass through to the underlying (Concurrent)HashMap

cdietrich: am not sure how expensive building closures is at thew caller site

cdietrich: and where should the pass go? there is no member in concurrent hashmap that accepts a supplier

szarnekow: Ok, I understand the point. Since the values are idempotent, it would be ok to compute the stuff twice for sure. But in general, a single call on the map would be preferable.

szarnekow: I was thinking about `java.util.Map.computeIfAbsent(K, Function<? super K, ? extends V>)` - you are right, it's not a supplier.

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
cdietrich: hmmm we dont store null there nor is the reader place the same as the writer so i am not sure if we need null at all

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
cdietrich: regarding string vs fqn (downstream api like getExoirtedObjects uses QN

szarnekow: Since it's a lossless conversion from QN to string but not the other way round, I'd prefer the cache to use strings nevertheless.

cdietrich: but wont that indicate to use the QN that has "more" information in order to keep it?

szarnekow: Maybe I misunderstood the semantics here.
Assuming I do have a class file name like `com.acme.Left$Right.class`. This could be the file that stems from a java source `class Left$Right { }` or `class Left { class Right{} }`. What would be the QN that is used from getExportedObjects? Would it be `[com, acme, Left, Right]` or `[com, acme, Left$Right]`? What we seem to put into the index is the latter `[com, acme, Left$Right]` - see XbaseQualifiedNameProvider, but we do seem to disagree on this, e.g. see JvmIdentifiableQualifiedNameProvider. If we can unambigously recreate the right names from the file names, the QN would be the better choice.

szarnekow: Question is: can we?

cdietrich: i dont know. i assue that jdt uses Outer$Inner in one source segment

cdietrich: => we create the QualifiedNames fromwhat jdt asks us

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
```diff
@@ -0,0 +1,159 @@
...
+	
+	def IBinaryType computeIfAbsent(QualifiedName qualifiedName, Function<? super QualifiedName, ? extends IBinaryType> fun) {
+		cache.computeIfAbsent(qualifiedName, fun)
+	}
+	
+	def void clear() {
+		cache.clear()
+	}
+	
+	static val NULL = new IBinaryType() {
```
cdietrich: am not sure if there is a cleaner way to do this

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.xtend
```diff
@@ -101,18 +104,34 @@ class JavaDerivedStateComputer {
...
 		if (data === null)
 			throw new IllegalStateException("no index installed")
 		// TODO use container manager
-		val nameEnv = new IndexAwareNameEnvironment(resource, classLoader, data, stubGenerator)
+		val nameEnv = new IndexAwareNameEnvironment(resource, classLoader, data, stubGenerator, classFileCache)
 		val compiler = new Compiler(nameEnv, DefaultErrorHandlingPolicies.proceedWithAllProblems(), resource.compilerOptions, [
+			for (cls : it.classFiles) {
+				// TODO What is with inner classes (they contain $)
+				// TODO is there a better way to obtain the class name
+				val key = qualifiedNameConverter.toQualifiedName(new String(cls.fileName).replace("/","."))
```
szarnekow: We probably do not need to dispatch through a QNConverter here. 
``` 
Splitter splitter = Splitter.on('/');
QualifiedName.create(splitter.splitToList(CharBuffer.wrap(fileName))
```
This would avoid repetitive string allocations and s&r operations.

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -51,9 +62,22 @@ import org.eclipse.emf.ecore.resource.Resource
...
+				//TODO is that ok
+				classFileCache.put(className, null)
 				return null;
 			}
-			val reader = ClassFileReader.read(url.openStream, fileName)
+			// todo: try with resources/ close Stream
+			var InputStream stream = null
+			val IBinaryType reader = try {
+				stream = url.openStream
+				ClassFileReader.read(url.openStream, fileName)
```
cdietrich: @szarnekow do we have a util that mimics try with resources in xtend?

cdietrich:
- have created 2.16-SNAPSHOT play branches https://github.com/eclipse/xtext-lib/tree/cd_fake216 and https://github.com/eclipse/xtext-core/tree/cd_fake216 that build against 2.15-SNAPSHOTs and this branch: 
- here is the job http://services.typefox.io/open-source/jenkins/job/xtext-core/job/cd_fake216/
- as you can see build times have basically halfed.

TODO:
- find clean solution
- answer TODOs in the code
- develop a concept for cleanup (maybe we should introduce a interface for the adapter and have a cleanup phase e.g. in incremental builder)
- manage discarded derived state.
- create better tests
- maybe we can "blow" up XtendIncrementalBuilderTest to work on more complex scenarios

gradle/bootstrap-setup.gradle
```diff
@@ -24,7 +24,7 @@ configurations {
 				if (requested.group == 'com.google.inject' && requested.name == 'guice')
 					useVersion(versions.guice)
 				if (requested.group == 'org.eclipse.jdt' && requested.name == 'org.eclipse.jdt.core')
-					useVersion(versions.jdtCore)
```
cdietrich: TODO: undo this change

gradle/upstream-repositories.gradle
```diff
@@ -25,7 +25,7 @@ repositories {
 	if (findProperty('useJenkinsSnapshots') == 'true') {
 		maven { url "http://services.typefox.io/open-source/jenkins/job/lsp4j/job/master/lastStableBuild/artifact/build/maven-repository/" }
 		maven { url jenkinsPipelineRepo('xtext-lib') }
-		maven { url jenkinsPipelineRepo('xtext-core') }
```
cdietrich: TODO: undo this change

gradle/versions.gradle
```diff
@@ -9,15 +9,15 @@ ext.versions = [
 	'xtext_bootstrap': '2.14.0',
 	'gradle_plugins': '0.1.0',
 	'xtext_gradle_plugin': '1.0.21',
-	'equinoxCommon' : '3.9.0',
```
cdietrich: TODO undo these changes

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ * @since 2.14
+ */
+@EmfAdaptable
+class ClassFileCache {
+	
+	static val NULL = new Object();
+	
+	//TODO: concurrency
```
cdietrich: remove TODOs

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
```diff
@@ -0,0 +1,66 @@
...
+		} else {
+			cache.put(qualifiedName, answer)
+		}
+	}
+	
+	def IBinaryType computeIfAbsent(QualifiedName qualifiedName, Function<? super QualifiedName, ? extends IBinaryType> fun) {
+		val result = cache.computeIfAbsent(qualifiedName) [
+			fun.apply(it) ?: NULL
+		]
+		if (result !== NULL) { // TODO should this ever happen
```
cdietrich: this way the commen should go to the non null case, maybe completely removed

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -50,13 +62,31 @@ import org.eclipse.emf.ecore.resource.Resource
...
+			var InputStream stream = null
+			val IBinaryType reader = try {
+				stream = url.openStream
+				ClassFileReader.read(stream, fileName)
+				//TODO what if read fails
+			} finally {
+				if (stream !== null) {
+					stream.close
+				}
+			}
```
cdietrich: removes TODOs

szarnekow: Done

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
szarnekow: Done

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/ClassFileCache.xtend
szarnekow: Done

PR #286, opened 2018-08-13
Cd issue261 c

cdietrich:
@szarnekow here a variant with concurrent hashmap, null object and strings instead of fqn.
am not sure if i like it


org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.xtend
```diff
@@ -12,30 +12,41 @@ import org.eclipse.xtext.common.types.descriptions.EObjectDescriptionBasedStubGe
...
+	Map<String, NameEnvironmentAnswer> cache = newHashMap()
     
 	override cleanup() {
 		cache.clear
 	}
 
 	override findType(char[][] compoundTypeName) {
 		val className = QualifiedName.create(compoundTypeName.map[String.valueOf(it)])
-		return findType(className)
+		return findType(className.toString)
```
szarnekow: No need to convert it to a QN in the first place.

cdietrich: how to easily do the join then?

cdietrich: list.join(".") ?

cdietrich: i will add the concurrent hashmap to the other one

szarnekow: org.eclipse.jdt.core.compiler.CharOperation.toString(char[][])

cdietrich: we should decide to go for one of the two appoaches

cdietrich:
@szarnekow is it ok if we close this one and follow the QN approach in the other pr?

PR #301, opened 2018-09-15
Make the "Cannot root twice" error message more explicit

gallandarakhneorg:
Sometimes, when the JVM model inferrer contains code that leads to a "Cannot root twice", it is helpful to know that is the Eobject that the system tries to root twice times.
This patch add the textual description of the EObject in the error message.


kthoms:
This commit is not signed. Unfortunately contributing with the GitHub web interface is still not supported properly.

kthoms:
Besides that LGTM

gallandarakhneorg:
Commit and PR are now signed.

PR #302, opened 2018-09-18
[https://github.com/eclipse/xtext-xtend/issues/554]

miklossy:
Ensure static and final modifiers generated in the right order.

- Modify the JvmModelGenerator to generate the static and the final
modifiers in the right order (corresponding to the java language
specification) 'static final' instead of 'final static'.


kthoms:
Thx @miklossy !

PR #305, opened 2018-09-23
[eclipse/xtext#1272] Set Xtext bootstrap 2.15

kthoms:
Changed target definition.


cdietrich:
Arggg the signature problem workaround is still in place

PR #323, opened 2018-10-18
Change the visibility of CandidateComparisonResult to public.

gallandarakhneorg:
It is impossible for the subtypes of `AbstractPendingLinkingCandidate` to call the `compareTo` function because the visibility of `CandidateComparisonResult` is package.
The visiblity should be changed to `public` for enabling the subtypes' functions to call `compareTo`.


cdietrich:
thx @gallandarakhneorg

PR #324, opened 2018-10-19
Support JvmCustomAnnotationValue when testing constantExpression.

gallandarakhneorg:
When the @Inline annotation is used into a Xtend or DSL code, and the
field "constantExpression" is specified into this annotation, the Xbase
compiler fails with a class cast exception.

This issue is due to the fact that the value associated "constantExpression" is not of type
JvmBooleanAnnotationValue, but of type JvmCustomAnnotationValue (with a
XBooleanLiteral inside).

This PR updates the `isConstantExpression` function in order to support both cases.


cdietrich:
i wonder if we can write a unit test for that

cdietrich:
e.g. something like CompilerXtendIssue256Test in xtend

gallandarakhneorg:
See https://github.com/eclipse/xtext-xtend/pull/615

cdietrich:
thx @gallandarakhneorg

PR #328, opened 2018-11-05
[eclipse/xtext-xtend#620] removed issue data to allow better multi quickfix

cdietrich:
[eclipse/xtext-xtend#620] removed issue data to allow better multi quickfix


cdietrich:
not needed

PR #332, opened 2018-11-14
[#1168] Minimal target platform Oxygen.3a

kthoms:
Update lower bound version constraints on platform related bundles with
versions from an Oxygen.3a platform.


org.eclipse.xtext.common.types/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,7 @@ Require-Bundle:
  org.eclipse.emf.mwe2.lib;bundle-version="2.2.0";resolution:=optional;x-installation:=greedy,
  org.objectweb.asm;bundle-version="[7.0.0,7.1.0)",
  org.eclipse.xtend.lib,
- com.google.guava;bundle-version="[14.0.0,22.0.0)"
+ com.google.guava;bundle-version="[15.0.0,22.0.0)"
```
cdietrich: am not sure if we want this already. orbit does not contain 15 for a long time but we have the 14-22) range all over the place, in gradle/maven too

kthoms: updated.

PR #341, opened 2018-11-21
[eclipse/xtext-core#967] [Generator] Add option to set GenModel.rootExtendsClass from EMFGeneratorFragment

cdietrich:
[eclipse/xtext-core#967] [Generator] Add option to set GenModel.rootExtendsClass from EMFGeneratorFragment


org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/ecore/EMFGeneratorFragment.java
```diff
@@ -148,6 +148,8 @@
 	
 	private boolean bindEPackageAndEFactory = false;
 	
+	private String rootExtendsClass = "org.eclipse.emf.ecore.impl.MinimalEObjectImpl$Container";
```
szarnekow: Maybe make this protected or add a protected getter in case a subclass exists.

PR #342, opened 2018-11-21
Set reference kind when we know it should be set

tivervac:
see eclipse/xtext-core#965

### DIFF
--- a/org.eclipse.xtext.xbase.ide/src/org/eclipse/xtext/xbase/ide/contentassist/XbaseIdeCrossrefProposalProvider.xtend
+++ b/org.eclipse.xtext.xbase.ide/src/org/eclipse/xtext/xbase/ide/contentassist/XbaseIdeCrossrefProposalProvider.xtend
@@ -45,6 +45,7 @@ class XbaseIdeCrossrefProposalProvider extends IdeCrossrefProposalProvider {
 			val converter = getTypeConverter(context.resource)
 			val objectOrProxy = candidate.getEObjectOrProxy
 			return proposalCreator.createProposal(proposalString, context) [ result |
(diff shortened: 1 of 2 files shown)


tivervac:
@cdietrich Might want to wait for master to build to see we didn't break anything with eclipse/xtext-core#965

org.eclipse.xtext.xbase.ide/src/org/eclipse/xtext/xbase/ide/contentassist/XbaseIdeCrossrefProposalProvider.xtend
```diff
@@ -44,7 +44,7 @@ class XbaseIdeCrossrefProposalProvider extends IdeCrossrefProposalProvider {
 					0
 			val converter = getTypeConverter(context.resource)
 			val objectOrProxy = candidate.getEObjectOrProxy
-			return proposalCreator.createProposal(proposalString, context) [ result |
+			val proposal = proposalCreator.createProposal(proposalString, context) [ result |
```
cdietrich: Any reason not to move this to the closure?

cdietrich: Am not sure if the create proposal can return null if prefix is bad

tivervac: My fault, didn't notice that this was an initialisation closure. If I do it in the closure, it's nullsafe.
Will replace my current implementation for the one you proposed.

PR #345, opened 2018-11-22
Set reference kind when we know it should be set

tivervac:
Follow up from eclipse/xtext-web#82


cdietrich:
lets see what ci says. does this affect web again?

tivervac:
@cdietrich It does, see eclipse/xtext-web#83

cdietrich:
build looks good

cdietrich:
build looks good

PR #352, opened 2018-12-13
[eclipse/xtext#1322] switched to tycho 1.3.0

cdietrich:
[eclipse/xtext#1322] switched to tycho 1.3.0

### DIFF
--- a/releng/p2/pom.xml
+++ b/releng/p2/pom.xml
@@ -13,7 +13,6 @@
 	</parent>
 
 	<properties>
-		<tycho-version>1.2.0</tycho-version>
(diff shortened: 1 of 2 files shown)


releng/p2/pom.xml
```diff
@@ -13,7 +13,6 @@
 	</parent>
 
 	<properties>
-		<tycho-version>1.2.0</tycho-version>
```
szarnekow: Wird die geerbt?

cdietrich: it always was. the p2 generator gradle plugin we used before was just too stupid to know that.

PR #371, opened 2019-02-05
[#677] Make Constructor visible to make use of this class in other sc…

holgerschill:
…enarios


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/scoping/batch/StaticExtensionFeatureDescriptionWithImplicitFirstArgument.java
```diff
@@ -18,7 +18,10 @@
  */
 public class StaticExtensionFeatureDescriptionWithImplicitFirstArgument extends StaticExtensionFeatureDescription {
 
-	protected StaticExtensionFeatureDescriptionWithImplicitFirstArgument(
+	/**
+	 * @since 2.17
+	 */
+	public StaticExtensionFeatureDescriptionWithImplicitFirstArgument(
```
cdietrich: @szarnekow should we go this way or have a factory method?

szarnekow: I think it's fine this way. We should probably do this for the other discriptions, too.

PR #373, opened 2019-02-07
[eclipse/xtext#1274] Updated .classpath by Gradle

kthoms:
Changes done by Buildship 3.0.1.v20181217-1554


org.eclipse.xtext.common.types.tests/xtext.common.types.tests.launch
```diff
@@ -1,16 +1,17 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.jdt.junit.launchconfig">
```
cdietrich: ? why is this changed?

PR #378, opened 2019-02-10
Another round of classpath fixes

szarnekow:


gradle/eclipse-project-layout.gradle
```diff
@@ -76,6 +76,9 @@ eclipse {
 				if (source.kind == 'src' && source.path.endsWith('-gen') && !source.path.equals('xtend-gen') ) {
 					source.entryAttributes['ignore_optional_problems'] = 'true'
 				}
+				if (source.kind == 'output') {
```
cdietrich: @oehme is this a bug in buildship. without that line "bin/default" is picked which leads to wrong results

oehme: Not in Buildship, but possibly in Gradle. Can you open an issue with a small reproducer?

PR #385, opened 2019-03-01
[codegen] generateXbase

kthoms:


org.eclipse.xtext.xbase/plugin.properties
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2011 itemis AG (http://www.itemis.eu) and others.
```
szarnekow: This file may not be committed

kthoms:
Abort. Please do as needed later.

PR #390, opened 2019-03-07
[eclipse/xtext-xtend#395] Try with resources 

epoell:
Validate, that a given resources implements the AutoClosable Interface
and that mark as an error otherwise.
Compile correct Xtend code to Java code correctly, also if resource
definitions are more complex.
Asssume AutoClosable as the default type of a resource.


org.eclipse.xtext.xbase/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Xbase Model
 Bundle-Vendor: Eclipse Xtext
-Bundle-Version: 2.18.0.qualifier
+Bundle-Version: 2.17.0.qualifier
```
cdietrich: looks like this needs a rebase @epoell

epoell: In that MANIFEST I was now able to change the version.
But somehow I must have missed something at the rebase and it might still be the wrong version for all the other manifests, despite I have version 2.18 in all files locally...

org.eclipse.xtext.xbase/src-gen/org/eclipse/xtext/xbase/serializer/AbstractXtypeSyntacticSequencer.java
```diff
@@ -1,5 +1,9 @@
 /*******************************************************************************
+<<<<<<< HEAD
```
szarnekow: Merge problem

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
```diff
@@ -345,7 +346,7 @@ protected boolean hasJvmConstructorCall(XExpression obj) {
 	protected boolean needsSneakyThrow(XExpression obj, Collection<JvmTypeReference> declaredExceptions) {
 		IResolvedTypes resolvedTypes = getResolvedTypes(obj);
 		List<LightweightTypeReference> thrownExceptions = resolvedTypes.getThrownExceptions(obj);
```
szarnekow: The thrown exceptions should already include the exception declared on the AutoClosable.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -460,12 +461,55 @@ protected void _toJavaExpression(XBlockExpression expr, ITreeAppendable b) {
...
-		b.newLine().append("try {").increaseIndentation();
+
+		// Resources    defined before try-statement, for java versions < 7
+		if (isTryWithRes && !java7) {
+			for (XVariableDeclaration res : resources) {
+				b.newLine();
+				LightweightTypeReference type = appendVariableTypeAndName(res, b);
+				b.append(" = ");
+				if (res.getRight() instanceof XConstructorCall)
+					constructorCallToJavaExpression((XConstructorCall) res.getRight(), b);
```
szarnekow: The constructor invocation should happen in the try block. If the constructor defines an exception to be thrown, it can be caught by a catch of the try block.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -474,11 +518,17 @@ protected void _toJavaStatement(XTryCatchFinallyExpression expr, ITreeAppendable
...
 		}
 		b.decreaseIndentation().newLine().append("}");
-		appendCatchAndFinally(expr, b, isReferenced);
+
+		// Catch and Finally
+		appendCatchAndFinally(expr, b, isReferenced, java7);
 	}
 
-	protected void appendCatchAndFinally(XTryCatchFinallyExpression expr, ITreeAppendable b, boolean isReferenced) {
+	protected void appendCatchAndFinally(XTryCatchFinallyExpression expr, ITreeAppendable b, boolean isReferenced, boolean java7) {
```
szarnekow: Unfortunately we cannot change the signature of this method to stay backwards compatible

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -570,6 +625,14 @@ protected JvmTypeReference resolveSynonymType(JvmSynonymTypeReference reference,
 		LightweightTypeReference superType = lightweight.getOwner().getServices().getTypeConformanceComputer().getCommonSuperType(lightweight.getMultiTypeComponents(), lightweight.getOwner());
 		return superType.toJavaCompliantTypeReference();
 	}
+	
+	protected void appendFinallyWithResources(EList<XVariableDeclaration> resources, ITreeAppendable b) {
```
szarnekow: Please do not declare parameters of type EList. List should be sufficient.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/XbaseTypeComputer.java
```diff
@@ -576,6 +588,18 @@ protected void _computeTypes(XVariableDeclaration object, ITypeComputationState
 								null);
 						state.addDiagnostic(diagnostic);
 					}
+					if (isTryWithResources && (!variableType.isSubtypeOf(java.lang.AutoCloseable.class)
```
szarnekow: You could use AutoClosable.class instead of the FQN here, couldn't you?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/XbaseTypeComputer.java
```diff
@@ -544,7 +546,17 @@ protected void _computeTypes(XVariableDeclaration object, ITypeComputationState
 			ITypeComputationState initializerState = state.assignType(object, lightweightTypeReference).withExpectation(lightweightTypeReference);
 			initializerState.computeTypes(object.getRight());
 		} else {
-			ITypeComputationState initializerState = lightweightTypeReference != null ? state.withExpectation(lightweightTypeReference) : state.withNonVoidExpectation();
+			ITypeComputationState initializerState;
+			if (isTryWithResources) {
+				initializerState = (lightweightTypeReference != null)
+						? state.withExpectation(lightweightTypeReference)
+						: state.withExpectation(getRawTypeForName(java.lang.AutoCloseable.class, state));
```
szarnekow: Same here

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
cdietrich: I wonder if we should drop support for older targets anyway

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
epoell: Indeed.
However when the resource is defined with a lambda expression, it is assumed to be an AutoClosable, but the corresponding exceptions are not listed in resolvedTypes.getThrownExceptions(obj). For that case I implemented the additional method.

Should I refine the method to that specific case, that the right hand side of the resource definition is a lambda expression?


example:
```
try (var r = [System.out.println("Closing")]) {} 
```
would generate the Java Code:
```
try (AutoCloseable r = ((AutoCloseable) () -> {
   System.out.println("Closing");
})) {}
```
and is marked as error, that AutoClosable throws an unhandled excepetion.
With the additional method is would result in
```
try {
   try (AutoCloseable r = ((AutoCloseable) () -> {
      System.out.println("Closing");
   })) {
   }
} catch (Throwable _e) {
   throw Exceptions.sneakyThrow(_e);
}
```

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
epoell: it's changed, it is not necessary to pass that as a parameter.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/AbstractXbaseCompiler.java
szarnekow: It should collect all the declared exceptions of the close methods of the resource vars

epoell: I has some more thoughts on and experiments with this issue:

Indeed, all exceptions get collected, at least if `close()` is explicitly called within the try's expression block.
However (independent from if the rhs is a lambda expression or not), if the resource is considered a not further specified `AutoClosable` and if `close()` is not explicitly called, but implicitly by the try with resource statement, the java error is thrown, that the resource has unhandled Excpetions. See cases 1 and 2 below.

I'm not sure if I can find out how to convince the collector to collect an unhandled thrown excepetion in case the resource is an AutoClosable. The much easier but less clean solution is to stick to the initial attempt, using a method to check if the resource is an AutoClosable and define that it needs a SneakyThrow in that case.

Case 1:
The Xtend Code with `someClosable` as an AutoClosable does not call any methods with the expression block and thus is considered to not throw any exceptions. 
```
val reader1 = new StringReader("Text1")
val reader2 = new StringReader("Text2")
val array = #[reader1,reader2]
		
try (val AutoCloseable someClosable = array.get(0)){}
```

In the generated Java Code the error is marked: "Unhandled exception type Exception thrown by automatic close() invocation on someClosable"
```
final StringReader reader1 = new StringReader("Text1");
final StringReader reader2 = new StringReader("Text2");
final List<StringReader> array = Collections.<StringReader>unmodifiableList(CollectionLiterals.<StringReader>newArrayList(reader1, reader2));
try (final AutoCloseable someClosable = array.get(0)) {
}
```

Case 2:
If one calls `someClosable.close()` explicitly in the try expression, in the generated Java Code an Excpetion is catched:
```
try {
      final StringReader reader1 = new StringReader("Text1");
      final StringReader reader2 = new StringReader("Text2");
      final List<StringReader> array = Collections.<StringReader>unmodifiableList(CollectionLiterals.<StringReader>newArrayList(reader1, reader2));
      try (final AutoCloseable closableOhne = array.get(0)) {
        closableOhne.close();
      }
} catch (Throwable _e) {
      throw Exceptions.sneakyThrow(_e);
}
```

Case 3:
If someClosable is a StringReader no exception on close is thrown and thus a try withour a cacth does not have any errors. If one would call e.g. `read()` in the expression block, a catch clause would be generated.
Xtend Code:
```
try (val someClosable = new StringReader("Text")){}
```
JavaCode without error, even though not catch clause:
```
try (final StringReader someClosable = new StringReader("Text")) {
}
```

Case 4:
If someClosable is another implementation of AutoClosable, like FileInputStream, FileReader, etc, a catch clause is generated as well.
Xtend Code:
```
try (val someCloseable = new FileInputStream("test.txt")) {}
```
Java Code:
```
try {
    try (final FileInputStream someCloseable = new Function0<FileInputStream>() {
        public FileInputStream apply() {
          try {
            return new FileInputStream("test.txt");
          } catch (Throwable _e) {
            throw Exceptions.sneakyThrow(_e);
          }
        }
      }.apply()) {
      }
} catch (Throwable _e) {
      throw Exceptions.sneakyThrow(_e);
}
```

Thus I think it is legit to call for a sneakyThrow whenever a resource is an AutoClosable. It also does not hurt if it actually isn't needed (as it should be in case 2).

epoell: One could add a query in `org.eclipse.xtext.xbase.typesystem.util.ThrownExceptionSwitch.caseXTryCatchFinallyExpression(XTryCatchFinallyExpression)`, where the catch clauses of a try statement are processed. In case of an AutoClosable (or XClosure) resource, an exception would be added to the list of thrown exceptions.
But I'm not convinced, if this attempt is cleaner than the simple one above.
And I'm not quite sure, how to add the exception.

cdietrich:
this pr was replaced by https://github.com/eclipse/xtext-extras/pull/403

cdietrich:
thx for your contribution @epoell

PR #396, opened 2019-03-25
mark 'Abstract*' class as abstract

cdietrich:
mark 'Abstract*' class as abstract


cdietrich:
@szarnekow what do you think?

PR #407, opened 2019-04-04
[xtext-xtend#734] Fixed restore logic of runtime injector providers

szarnekow:


cdietrich:
There was an explicit decision by the team not to adapt the old generator

PR #411, opened 2019-04-11
Protected visibility for convertWrapperToPrimitive() and convertPrimitiveToWrapper()

gallandarakhneorg:
In order to allow overriding, the functions `convertWrapperToPrimitive()` and `convertPrimitiveToWrapper()` should have a protected accessibility.

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/TypeConvertingCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/TypeConvertingCompiler.java
@@ -38,6 +38,7 @@ import com.google.common.collect.Lists;
 
 /**
  * @author Sven Efftinge - Initial contribution and API
+ * @author Stephane Galland - Change visibilities of convertion functions.
  */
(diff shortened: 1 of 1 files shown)


cdietrich:
@gallandarakhneorg do you have a specific usecase in mind

gallandarakhneorg:
Yes, I have a use case for `convertWrapperToPrimitive()`: generating null-safe java code.

One example is the conversion for a `Byte` to a `byte` that generates `(v == null ? 0 : v.byteValue())`.
A second example is the conversion from `Boolean` to `boolean` that generates `(v == null ? false : v.booleanValue())`.

The current implementation of `convertWrapperToPrimitive()` does not enforce the null-safe behavior.

I have not specific use case for `convertPrimitiveToWrapper()`. I have assumed that if one function is protected, the other should also be.

Because the null-safe behavior is not part of the Java operational semantic, and the Xbase follows this operational semantics, I think it is mandatory to allow DSL developers to override this semantic in subclasses.

szarnekow:
@cdietrich what do you think?

cdietrich:
thx @gallandarakhneorg

Issue #415, opened 2019-04-18
Improve exception handling for try-with-resource


szarnekow:
The Java6-style code generator and the interpreter may suffer from unclosed resources, if the close method throws an exception.

```
try (val from = new FileInputStream(..); val to = new FileOutputStream(..)) {
  // do stuff
}
```
If `to.close()` throws, `from.close()` must still be invoked and we must propagate the exception of `to.close()` afterwards.

szarnekow:
Hey @max-leuthaeuser, maybe you want to give this a shot?

szarnekow:
Thank you, @max-leuthaeuser

Issue #244, opened 2018-03-24
Develop a concept to support "new" features with xbase


cdietrich:
Develop a concept to support "new" features like try with resources, try multi catch, more number literals and operators with xbase with or without breaking clients e.g. by introducing a new subgrammar similar to XbaseWith Annotations

szarnekow:
Hi @loradd that's an unfortunate overlap since we do already have this first draft of the implementation locally. Maybe you want to share your solution so we can discuss how to join forces, though.

loradd:
Hi @szarnekow, sure! 

I have first defined a new `XTernaryIfExpression` rule "between" `XExpression` and `XAssignment` in ` org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/Xbase.xtext`, as follows:

    XExpression returns XExpression:
        XTernaryIfExpression
    ;

    XTernaryIfExpression returns XExpression:
        XAssignment =>({XIfExpression.if = current} '?' 
            then=XExpression (=>':' else=XExpression)?
        )?

Then, I changed the `XBaseSemanticSequencer` managing ternary expressions as ordinary if expressions in `org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/serializer/XbaseSemanticSequencer.java`, as follows:
    
    @Override
    protected void sequence_XIfExpression_XTernaryIfExpression(ISerializationContext context, XIfExpression semanticObject) {
        super.sequence_XIfExpression(context, semanticObject);
    }

loradd:
Yes, thank you!

cdietrich:
Closed as wont-fix. Feel free to reopen at github.com/eclipse/xtext

PR #416, opened 2019-04-23
[eclipse/xtext-extras#415] Improve exception handling for try-with-resource

max-leuthaeuser:
Invoke all close methods from resources in try-with-resources and
propagate thrown exceptions afterwards for the Java6-style code generator and for the interpreter.

Signed-off-by: Max Leuthäuser <max.leuthaeuser@itemis.de>

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
@@ -476,7 +476,7 @@ public class XbaseCompiler extends FeatureCallCompiler {
 		}
 		
 		String throwablesStore = null;
-		if (isTryWithResources) {
+		if (isTryWithResources && !nativeTryWithResources) {
 			// The try block and/or invoking the close method later on might throw.
 			// Hence, we collect those Throwables and propagate them later on.
 			throwablesStore = b.declareSyntheticVariable(Tuples.pair(expr, "_caughtThrowables"), "_ts");
@@ -493,7 +493,7 @@ public class XbaseCompiler extends FeatureCallCompiler {
 		if (isTryWithResources && !nativeTryWithResources) {
 			for (XVariableDeclaration res : resources) {
 				b.newLine();
-				// resource has to be declared and initialised with null before try
+				// resource has to be declared and initialized with null before try
 				// and with real AutoClosable implementation within try
 				// has to be var not val (final)
 				res.setWriteable(true);
@@ -556,6 +556,7 @@ public class XbaseCompiler extends FeatureCallCompiler {
 		final EList<XCatchClause> catchClauses = expr.getCatchClauses();
 		final XExpression finallyExp = expr.getFinallyExpression();
 		boolean isTryWithResources = !expr.getResources().isEmpty();
+		boolean nativeTryWithResources = isAtLeast(b, JAVA7); // If Java 7 or better: use Java's try with resources
 		
 		// Catch
 		if (!catchClauses.isEmpty()) {
@@ -573,7 +574,7 @@ public class XbaseCompiler extends FeatureCallCompiler {
 			}
 			b.append(" else {");
 			b.increaseIndentation().newLine();
-			if(isTryWithResources) {
+			if(isTryWithResources && !nativeTryWithResources) {
 				b.append(throwablesStore + ".add(" + variable + ");").newLine();
 			}			
 			appendSneakyThrow(expr, b, variable);
@@ -582,7 +583,6 @@ public class XbaseCompiler extends FeatureCallCompiler {
 		}
 
 		// Finally
-		boolean nativeTryWithResources = isAtLeast(b, JAVA7);
 		if (finallyExp != null || (!nativeTryWithResources && isTryWithResources)) {
 			b.append(" finally {").increaseIndentation();
 			if (finallyExp != null)
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
@@ -475,8 +475,21 @@ public class XbaseCompiler extends FeatureCallCompiler {
 			declareSyntheticVariable(expr, b);
 		}
 		
+		String throwablesStore = null;
+		if (isTryWithResources) {
+			// The try block and/or invoking the close method later on might throw.
+			// Hence, we collect those Throwables and propagate them later on.
+			throwablesStore = b.declareSyntheticVariable(Tuples.pair(expr, "_caughtThrowables"), "_ts");
+			b.newLine().append(List.class);
+			b.append("<").append(Throwable.class).append("> ");
+			b.append(throwablesStore);
+			b.append(" = new ");
+			b.append(ArrayList.class);
+			b.append("<").append(Throwable.class).append(">();");
+		}
+		
 		b.openPseudoScope();
-		// Resources    declared before the try-statement, for java versions < 7
+		// Resources declared before the try-statement, for java versions < 7
 		if (isTryWithResources && !nativeTryWithResources) {
 			for (XVariableDeclaration res : resources) {
 				b.newLine();
@@ -489,11 +502,11 @@ public class XbaseCompiler extends FeatureCallCompiler {
 				b.append(" = null;");
 			}
 		}
-
+				
 		// Try
 		b.newLine().append("try ");
 		
-		// Resources    for java versions >= 7
+		// Resources for java versions >= 7
 		// constructed with Java's try with resources
 		if (isTryWithResources && nativeTryWithResources) {
 			int isLast = resources.size();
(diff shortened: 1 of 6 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -644,11 +644,40 @@ protected JvmTypeReference resolveSynonymType(JvmSynonymTypeReference reference,
...
-		for (int i = resources.size() - 1; i >= 0; i--) {
-			XVariableDeclaration res = resources.get(i);
-			String resName = getVarName(res, b);
-			b.newLine().append("if (" + resName + " != null) ").append(resName + ".close();");
+	protected void appendFinallyWithResources(XTryCatchFinallyExpression expr, List<XVariableDeclaration> resources,
+			ITreeAppendable b) {
+		if (expr != null && !resources.isEmpty()) {
+			// Invoking the close method might throw. Hence, we collect those
+			// Throwables and propagate them later on.
+			String throwablesStore = b.declareSyntheticVariable(Tuples.pair(expr, "_catchedThrowables"), "_ts");
```
szarnekow: Catched -> caught

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -644,11 +644,40 @@ protected JvmTypeReference resolveSynonymType(JvmSynonymTypeReference reference,
...
+			b.newLine().append("java.util.List<Throwable> ");
```
szarnekow: Instead of using FQN in generated code, we should use types instead so that the import manager can take care of them

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -644,11 +644,40 @@ protected JvmTypeReference resolveSynonymType(JvmSynonymTypeReference reference,
...
+			b.append(throwablesStore);
+			b.append(" = new java.util.ArrayList<Throwable>();");
+			for (int i = resources.size() - 1; i >= 0; i--) {
+				XVariableDeclaration res = resources.get(i);
```
szarnekow: We could use openPseudoScope on the appendable to avoid unnecessary counters on the catch-vars in the generated code

max-leuthaeuser:
> This is a very good start. I would expect changes in the tests for Xtend, too. Is there a PR for these?

I will do this tomorrow. All other suggestions are handled in the latest commit.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
max-leuthaeuser: Done.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
max-leuthaeuser: Done.

max-leuthaeuser: Done.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
max-leuthaeuser: Done.

max-leuthaeuser:
Adapted Xtend tests are available at:
https://github.com/eclipse/xtext-xtend/pull/790

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -534,21 +548,21 @@ protected void _toJavaStatement(XTryCatchFinallyExpression expr, ITreeAppendable
 		b.decreaseIndentation().newLine().append("}");
 
 		// Catch and Finally
-		appendCatchAndFinally(expr, b, isReferenced);
+		appendCatchAndFinally(expr, b, isReferenced, throwablesStore);
 		b.closeScope();
 	}
 
-	protected void appendCatchAndFinally(XTryCatchFinallyExpression expr, ITreeAppendable b, boolean isReferenced) {
+	protected void appendCatchAndFinally(XTryCatchFinallyExpression expr, ITreeAppendable b, boolean isReferenced, String throwablesStore) {
```
szarnekow: In general this PR looks good to me. I'm not entirely sure though if we should change the signature of this method or lookup the throwablesStore from the appendable again. @cdietrich what do you think?

cdietrich: at least Jbase makes use of `appendCatchAndFinally` cc @LorenzoBettini

cdietrich: this i think keeping the api might be better. but on the other hand i assume lorenzo has to update jbase for `?:` and try-with-resources anyway

szarnekow: I do have a tendency towards API compatibility, too. So let's play this safe.

LorenzoBettini: I can also update Jbase according to the new API, though I'm also in favor of API compatibility ;)

what I probably missed is: conditional expression and try-with-resources are effectively part of Xbase? I seemed to understand that they were introduced in Xtend directly...

cdietrich: yes its xtend. i missed that. sorry

LorenzoBettini: So I take the chance: why not in Xbase? ;)

cdietrich: i assume this is https://github.com/eclipse/xtext-extras/issues/244 correct @szarnekow

szarnekow: Indeed it is. Changing Xbase.ecore is possible in a backwards compatible way whereas changing the grammar unfortunately is not (as of now)

cdietrich: but that means lorenzo could reuse ecore and co

szarnekow: Exactly. Reusing the compiler, the ecore, the validations, the interpreter etc is possible. It's only the grammar that needs to be adjusted.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -186,25 +187,35 @@ protected void _toJavaExpression(XListLiteral literal, ITreeAppendable b) {
 				if (literalType instanceof CompoundTypeReference) {
 					for (LightweightTypeReference c : literalType.getMultiTypeComponents()) {
 						if (c.isArray()) {
-							b.append("new ")
-							.append(c.getType()) // append raw type since we cannot create generic arrays
-							.append(" ");
+							b.append("new ").append(c.getType()) // append raw
+																	// type
```
szarnekow: Format all the things? What did happen here?

max-leuthaeuser: Damn. Eclipse auto-format fail.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -559,33 +576,40 @@ protected void appendCatchAndFinally(XTryCatchFinallyExpression expr, ITreeAppen
...
 			if (finallyExp != null)
 				internalToJavaStatement(finallyExp, b, false);
 			if (!nativeTryWithResources && isTryWithResources)
-				appendFinallyWithResources(expr.getResources(), b);
+				appendFinallyWithResources(expr, b);
 			b.decreaseIndentation().newLine().append("}");
 		}
 	}
 
+	protected void appendSneakyThrow(XTryCatchFinallyExpression expr, ITreeAppendable b, String variable) {
```
szarnekow: I'll add `@since 2.18`

szarnekow:
Thank you

Issue #445, opened 2019-06-14
Try-With-Resource broken with Java 8


cdietrich:
when running `org.eclipse.xtext.purexbase.test.XbaseIntegrationTest`
against java 8 target (we use 5 as default) it will fail
https://ci.eclipse.org/xtext/job/xtext-extras/job/cd_extras_issue113/lastCompletedBuild/testReport/org.eclipse.xtext.purexbase.test/XbaseIntegrationTest/testTryWithResources_finally/

the same problem can be reproduce with a simple xtend sample

```
package org.eclipse.xtext.purexbase.test

class Demo {

	def xxxx() {
		var testdata.ClosableWithList result
		try (var a = new testdata.ClosableWithList()) {
			a.add("body")
			result = a
		} finally {
			a.add("finally")
		}
		return result.printList
	}
}

```

```
public class Demo {
  public String xxxx() {
    ClosableWithList result = null;
    try (ClosableWithList a = new ClosableWithList()) {
      a.add("body");
      result = a;
    } finally {
      a.add("finally");
    }
    return result.printList();
  }
}
```

a cannot be resolved

cdietrich:
=> with a should not be in scope in finally or we should rethink compilation strategy
cc @szarnekow

szarnekow:
`a` must not be reachable in the finally expression. I'll look at it.

cdietrich:
maybe we should have a look at the problems i have with eclipse/xtext#2344 to avoid such things for the future

szarnekow:
Fixed

cdietrich:
Fixed in 2.19

PR #449, opened 2019-06-14
[#445] Fixed scoping in try-with-resources

szarnekow:


org.eclipse.xtext.purexbase.tests/src/org/eclipse/xtext/xbase/junit/evaluation/AbstractXbaseEvaluationTest.java
```diff
@@ -2928,13 +2928,12 @@
 	 */
 	@Test @Ignore public void testTryWithResources_catch() throws Exception {
```
cdietrich: btw why is here a ignore?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/XbaseTypeComputer.java
```diff
@@ -1166,17 +1166,18 @@ protected void _computeTypes(XTryCatchFinallyExpression object, ITypeComputation
 				caughtExceptions.add(referenceOwner.toLightweightTypeReference(catchClause.getDeclaredParam().getParameterType()));
 		}
 		
+		ITypeComputationState withExpectedExceptions = state.withExpectedExceptions(caughtExceptions);
```
cdietrich: do i see it correct we dont have a negative test here?

org.eclipse.xtext.purexbase.tests/src/org/eclipse/xtext/xbase/junit/evaluation/AbstractXbaseEvaluationTest.java
szarnekow: Xbase has no syntax for try with resource. It has to be enabled in sublangs for backwards compat reasons

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/computation/XbaseTypeComputer.java
szarnekow: I’ll add a linking test

PR #453, opened 2019-06-18
performance: cache system classpath in ClasspathScanner

cdietrich:

### DIFF
--- a/org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/descriptions/ClasspathScanner.xtend
+++ b/org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/descriptions/ClasspathScanner.xtend
@@ -52,16 +52,22 @@ class ClasspathScanner {
 		return ClassLoader.systemClassLoader.getDescriptors(true, packagePrefixes)
 	}
 	
+	var String[] systemClasspath = null
+	
 	def String[] getSystemClasspath() {
+		if (systemClasspath !== null) {
+			return systemClasspath;
+		}
 		try (val ScanResult scanResult = new ClassGraph()
 				.enableSystemJarsAndModules()
 				.addClassLoader(ClassLoader.getSystemClassLoader())
 				.scan()) {
 			val classpathURIs = scanResult.classpathURIs
-			return classpathURIs.map[path].filterNull
+			systemClasspath = classpathURIs.map[path].filterNull
+			return systemClasspath;
 		}
 	}
-
+	
 	protected def Iterable<ITypeDescriptor> loadDescriptors(ClassLoader classLoader, boolean bootstrap,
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/descriptions/ClasspathScanner.xtend
```diff
@@ -52,16 +52,22 @@ class ClasspathScanner {
 		return ClassLoader.systemClassLoader.getDescriptors(true, packagePrefixes)
 	}
 	
+	var String[] systemClasspath = null
```
kthoms: Could this variable be accessed by different threads? Is synchronization required?

kthoms: Maybe just a theoretical issue. When 2 threads have null for that field, both would do the same computation and last wins. So if you think that does not matter then I'm fine with that.

szarnekow: What bugs me more is that we cache an array which is by definition not immutable. Callers may change the entries.

cdietrich: the array is in the signature

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/descriptions/ClasspathScanner.xtend
```diff
@@ -52,16 +52,22 @@ class ClasspathScanner {
...
+	
 	def String[] getSystemClasspath() {
+		if (systemClasspath !== null) {
+			return systemClasspath;
```
szarnekow: `return systemClasspath.clone()` ?

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/descriptions/ClasspathScanner.xtend
```diff
@@ -52,16 +52,22 @@ class ClasspathScanner {
...
+		}
 		try (val ScanResult scanResult = new ClassGraph()
 				.enableSystemJarsAndModules()
 				.addClassLoader(ClassLoader.getSystemClassLoader())
 				.scan()) {
 			val classpathURIs = scanResult.classpathURIs
-			return classpathURIs.map[path].filterNull
+			systemClasspath = classpathURIs.map[path].filterNull
+			return systemClasspath;
```
szarnekow: Also here: clone may prevent harm.

PR #458, opened 2019-06-25
[eclipse/xtext-core#1183]: Fix tests on windows machines.

ArneDeutsch:
Signed-off-by: Arne Deutsch <Arne.Deutsch@itemis.de>

### DIFF
--- a/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
+++ b/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
@@ -3530,7 +3530,8 @@ public abstract class AbstractTypeProviderTest extends Assert {
 		String typeName = "com.google.common.collect.AbstractMultimap";
 		JvmGenericType type = (JvmGenericType) getTypeProvider().findTypeByName(typeName);
 		assertNotNull(type);
-		diagnose(type, "java:/Objects/javax.annotation.Nullable#javax.annotation.Nullable");
+		diagnose(type, "java:/Objects/javax.annotation.Nullable#javax.annotation.Nullable",
+				"java:/Objects/com.google.errorprone.annotations.CanIgnoreReturnValue#com.google.errorprone.annotations.CanIgnoreReturnValue");
 		Resource resource = type.eResource();
 		getAndResolveAllFragments(resource);
 		recomputeAndCheckIdentifiers(resource);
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
```diff
@@ -3530,7 +3530,8 @@ public void testFindTypeByName_AbstractMultimap() {
 		String typeName = "com.google.common.collect.AbstractMultimap";
 		JvmGenericType type = (JvmGenericType) getTypeProvider().findTypeByName(typeName);
 		assertNotNull(type);
-		diagnose(type, "java:/Objects/javax.annotation.Nullable#javax.annotation.Nullable");
+		diagnose(type, "java:/Objects/javax.annotation.Nullable#javax.annotation.Nullable",
```
cdietrich: @ArneDeutsch this seem to be unrelated and caused by different classpaths.
thus i assume this will fail with gradle

szarnekow: Maybe we should exclude errorprone in the pom?

cdietrich: am not sure where this is coming from on arnes machine

cdietrich: will test this on my windows machine in the evening

cdietrich: @szarnekow i dont understand the test. there is a guard
`if (diagnostic.getSeverity() != Diagnostic.OK) {` in   `org.eclipse.xtext.common.types.access.impl.AbstractTypeProviderTest.diagnose(EObject, String...)  `
so that if there is no diagnostic at all the test will work regardless if there are any expected diagnostics or not.
is this behaviour intended? if no why does it behave differently on windows (or does it on your branch only @ArneDeutsch )

szarnekow: In different classpath configurations, we do have different missing transitive types. `diagnose(EObject, String...)` is a poor attempt to whitelist missing types.

szarnekow: It would be great if we could get rid of this.

ArneDeutsch:
Not ready.

PR #473, opened 2019-07-12
use lastStableBuild and not lastSuccessfulBuild

cdietrich:
use lastStableBuild and not lastSuccessfulBuild


cdietrich:
@Hejado please review

PR #522, opened 2019-11-15
moved GenerateEMF to a non production area

cdietrich:
moved GenerateEMF to a non production area


cdietrich:
@szarnekow i have already fixed that on master

PR #531, opened 2019-11-22
avoid unwanted build.properties changes

cdietrich:


cdietrich:
Need to do more checks why this happens in the first place 
Maybe bad gitignore

cdietrich:
looks like the default mechanic is buggy when the generated folder is not model/generated

Issue #538, opened 2019-12-03
Java Source Lang causes NPE


cdietrich:
```
Caused by: java.lang.NullPointerException
        at org.eclipse.jdt.internal.compiler.CompilationResult.<init>(CompilationResult.java:101)
        at org.eclipse.jdt.internal.compiler.Compiler.internalBeginToCompile(Compiler.java:829)
        at org.eclipse.jdt.internal.compiler.Compiler.beginToCompile(Compiler.java:394)
        at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:444)
        at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:426)
        at org.eclipse.xtext.java.resource.JavaDerivedStateComputer.installFull(JavaDerivedStateComputer.java:279)
        at org.eclipse.xtext.java.resource.JavaResource.lambda$installFull$1(JavaResource.java:195)
        at org.eclipse.xtext.java.resource.JavaResource.initializing(JavaResource.java:205)
        at org.eclipse.xtext.java.resource.JavaResource.installFull(JavaResource.java:199)
        at org.eclipse.xtext.java.resource.JavaResource.getContents(JavaResource.java:168)
        at org.eclipse.xtext.build.IncrementalBuilder$InternalStatefulIncrementalBuilder.lambda$launch$2(IncrementalBuilder.java:264)
```

`org.eclipse.jdt.internal.compiler.Compiler.compile(ICompilationUnit[])`
is called with nulls in the list

cdietrich:
for some reasons we have java resources with compilation unit null

cdietrich:
![Bildschirmfoto vom 2019-12-03 08-38-19](https://user-images.githubusercontent.com/238736/70029864-4edcbe80-15a8-11ea-9228-b9ea919aba53.png)

cdietrich:
workaround is not having `package-info.java`s

szarnekow:
Probably also true for module-info

cdietrich:
org.eclipse.xtext.java.resource.JavaDerivedStateComputer.isInfoFile(Resource)
might be missing at some places

cdietrich:
here is a unit test

```
    @Test def void testPackageInfo() {
        val rs = resourceSet( 'a/package-info.java' -> '''
            package a;
        ''', 'a/MyClass.java' -> '''
        	package a;
        	public class MySuperClass {
        	}
        ''')
        for (r : rs.resources) {
[...]

PR #542, opened 2019-12-04
[#538] add special treatment for module and package info on unloading too

cdietrich:
[#538] add special treatment for module and package info on unloading too

### DIFF
--- a/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.xtend
+++ b/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.xtend
@@ -46,7 +46,8 @@ class JavaDerivedStateComputer {
 			unloader.unloadRoot(eObject) 
 		}
 		resource.getContents().clear
-		
+		if (resource.isInfoFile)
+			return;
 		if (resource.compilationUnit !== null) {
 			val classFileCache = resource.resourceSet.findOrCreateClassFileCache
 			classFileCache.addResourceToCompile(resource)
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.xtend
```diff
@@ -46,7 +46,8 @@ class JavaDerivedStateComputer {
 			unloader.unloadRoot(eObject) 
 		}
 		resource.getContents().clear
-		
+		if (resource.isInfoFile)
```
szarnekow: If I understand this correctly, JavaResources for info files do have a strong reference to a compilation unit.
Should that one be released, too? Assuming we would never have that reference, the null check below would have been sufficient. Other code here could be simplified, too.

cdietrich: yes and no. i dont see where this reference would be installed in the first place.
or do you mean in other cases?

szarnekow: I was thinking about a special treatment a little earlier in the lifecycle, e.g. in `JavaResource.doLoad(InputStream, Map<?, ?>)`

cdietrich: i am not sure if that would have any unwanted side effects

szarnekow: True. That's also the reason why I approved this minimally invasive fix here.

PR #584, opened 2020-04-10
[eclipse/xtext-xtend#1035] Missing validation in context of static interface methods

szarnekow:
[eclipse/xtext-xtend#1035] Missing validation in context of static interface methods


cdietrich:
can you check the downstream test failures?

szarnekow:
I didn’t adjust the tp downstream. Probably the reason. Yeah split repo. I’ll look at this later

PR #599, opened 2020-05-05
[eclipse/xtext#1738] avoid usage of icu.Transliterator

cdietrich:
[eclipse/xtext#1738] avoid usage of icu.Transliterator


org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/grammarAccess/UnicodeCharacterDatabaseNames.java
```diff
@@ -13,6 +13,8 @@
 /**
  * @author Moritz Eysholdt
  * @author Sven Efftinge
+ * 
+ * @deprecated will be removed with Xtext 2.23
  */
 @Deprecated
 public class UnicodeCharacterDatabaseNames {
```
szarnekow, line 20: The deprecated impl could delegate to Character.getName(..), too such that we can remove the icu dependency earlier?

PR #609, opened 2020-05-21
GenerateXbaseTestLanguages: setting the generateEclipse flag to true.

miklossy:


miklossy:
@cdietrich As discussed offline, I created the PR to modify the `GenerateXbaseTestLanguages.mwe2 `workflow to activate the eclipse ui generation by default. I am convinced that this setting will help to avoid further inconveniences in the future.

PR #633, opened 2020-08-01
[eclipse/xtext#1777] ported xtend code to java

cdietrich:
[eclipse/xtext#1777] ported xtend code to java


org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/adapter/FragmentAdapter.java
```diff
@@ -0,0 +1,626 @@
...
+			}
+		}
+		return false;
+	}
+
+	public IGeneratorFragment getFragment() {
+		return fragment;
+	}
+
+	public void setFragment(IGeneratorFragment fragment) {
```
szarnekow, line 615: Cosmetics: Can be annotated with `@Mandatory`

PR #638, opened 2020-08-17
fixed warnings when compiling against Java 11

cdietrich:
fixed warnings when compiling against Java 11


cdietrich:
@szarnekow i am not sure why there was "new" used in the first place. maybe you can remember

PR #652, opened 2020-10-12
Cleanup XFunctionTypeRefs

nbhusare:
Signed-off-by: nbhusare <neerajbhusare@gmail.com>

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xtype/util/XFunctionTypeRefs.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xtype/util/XFunctionTypeRefs.java
@@ -8,6 +8,9 @@
  */
 package org.eclipse.xtext.xtype.util;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.InternalEObject;
@@ -19,6 +22,7 @@ import org.eclipse.xtext.common.types.JvmTypeReference;
 import org.eclipse.xtext.common.types.JvmVoid;
 import org.eclipse.xtext.common.types.TypesFactory;
 import org.eclipse.xtext.common.types.access.impl.ClassURIHelper;
+import org.eclipse.xtext.common.types.access.impl.URIHelperConstants;
 import org.eclipse.xtext.xbase.lib.Functions;
 import org.eclipse.xtext.xbase.lib.Procedures;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xtype/util/XFunctionTypeRefs.java
```diff
@@ -19,25 +19,29 @@
...
 	public static String buildUri(boolean procedure, int functionParamCount) {
 		int paramCount = Math.min(6, functionParamCount);
 		if (procedure) {
-			return "java:/Objects/" + Procedures.class.getCanonicalName() + "#" + Procedures.class.getCanonicalName()
-					+ "$Procedure" + Integer.valueOf(paramCount);
+			return classURIHelper.getFullURI(Procedures.class).toString() + "$Procedure" + paramCount;
 		}
-		return "java:/Objects/" + Functions.class.getCanonicalName() + "#" + Functions.class.getCanonicalName()
-				+ "$Function" + Integer.valueOf(paramCount);
+		return classURIHelper.getFullURI(Functions.class).toString() + "$Function" + paramCount;
```
szarnekow, line 44: I'm not sure that converting this to a URI and subsequently calling toString again is a good tradeoff. 
You may want to replace the literal `"java:/Objects/"` by a concatenation from URIHelperConstants, though.

Since `buildURI` is only used from within this class, it could also be an option to overhaul the flow and do something based on Class.forName instead or even keep an immutable list of the 2x6 allowed combinations such that we can return the URI in constant time without any string / URI manipulation.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xtype/util/XFunctionTypeRefs.java
```diff
@@ -19,25 +19,29 @@
...
 import org.eclipse.xtext.xbase.lib.Procedures;
 
 /**
  * @author Anton Kosyakov - Initial contribution and API
  */
 public class XFunctionTypeRefs {
+	
+	private static final Primitives primitives = new Primitives();
+
+	private static final ClassURIHelper classURIHelper = new ClassURIHelper();
```
szarnekow, line 33: You may want to use this classURIHelper in `computeTypeUri`, too.

cdietrich:
please check for test fails

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xtype/util/XFunctionTypeRefs.java
nbhusare: Could you please elaborate on the second point above?

nbhusare: > even keep an immutable list of the 2x6 allowed combinations such that we can return the URI in constant time without any string / URI manipulation.

@szarnekow Updated the code. The method `buildUri` returns a URI in constant time.

nbhusare:
@cdietrich I had reverted the code in the method `getWrappedType`.  The injected variable `TypeReferences typeReferences` in the `Primitives` was not getting initialized as I was using `new Primitives()`. Even if I use injection, I cannot use it in the `getWrappedType` method.

szarnekow:
Code looks good. One question though: Why did you decide to touch this?

nbhusare:
@szarnekow It was suggested by @cdietrich

nbhusare:
@szarnekow Please review and approve.

szarnekow (approved):
Approved. Thank you for the cleanup. Please go ahead with the merge.

PR #657, opened 2020-10-19
fixed more junit deprecations

cdietrich:
fixed more junit deprecations


kthoms (approved):
I usually prefer static imports for assertion methods. If you want you could change that, or leave as it is.

PR #674, opened 2020-12-01
bugfix 'return void' in XbaseInterpreter

asheiduk:
Evaluating the following example with `XbaseInterpreter` yields an internal error:
```
def void foo(){
	return
}
def String bar(){
	foo();
	"bla"
}
bar()
```

The error is only shown in the console. The relevant stacktrace section is:
```
Caused by: java.lang.IllegalArgumentException: Unhandled parameter types: [null, org.eclipse.xtext.xbase.interpreter.impl.DefaultEvaluationContext@3a687ba6, org.eclipse.xtext.builder.MonitorBasedCancelIndicator@7f6af28f]
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:271)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:204)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:285)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:252)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:204)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:447)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:228)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:204)
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.evaluate(XbaseInterpreter.java:190)
	at org.eclipse.xtext.mqrepl.interpreter.ModelQueryLanguageInterpreter.invokeOperation(ModelQueryLanguageInterpreter.java:58)
```

The patch does not assume that `return` always has an expression which can be given `internalEvaluate`.

Signed-off-by: Andreas Heiduk <asheiduk@gmail.com>

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
@@ -294,8 +294,13 @@ public class XbaseInterpreter implements IExpressionInterpreter {
 	}
 	
 	protected Object _doEvaluate(XReturnExpression returnExpr, IEvaluationContext context, CancelIndicator indicator) {
-		Object returnValue = internalEvaluate(returnExpr.getExpression(), context, indicator);
-		throw new ReturnValue(returnValue);
+		XExpression expression = returnExpr.getExpression();
+		if (expression != null) {
+			Object returnValue = internalEvaluate(returnExpr.getExpression(), context, indicator);
+			throw new ReturnValue(returnValue);
+		} else {
+			throw new ReturnValue(null);
+		}
 	}
 
 	/**


cdietrich:
wonder if we should add a testcase to
AbstractXbaseEvaluationTest / XbaseInterpreterTest

cdietrich:
@asheiduk please make sure you have a current ECA signed https://www.eclipse.org/legal/ECA.php

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
```diff
@@ -294,8 +294,13 @@ protected Object _doEvaluate(XNullLiteral literal, IEvaluationContext context, C
 	}
 	
 	protected Object _doEvaluate(XReturnExpression returnExpr, IEvaluationContext context, CancelIndicator indicator) {
-		Object returnValue = internalEvaluate(returnExpr.getExpression(), context, indicator);
-		throw new ReturnValue(returnValue);
+		XExpression expression = returnExpr.getExpression();
+		if (expression != null) {
+			Object returnValue = internalEvaluate(returnExpr.getExpression(), context, indicator);
```
cdietrich, line 299: wonder if this is a regression in internalEvaluate

asheiduk:
> @asheiduk please make sure you have a current ECA signed https://www.eclipse.org/legal/ECA.php
Done just now.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
cdietrich: and or if we should have an elseif for Null/Void in  org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XExpression, IEvaluationContext, CancelIndicator)

cdietrich: @szarnekow what do you think?

szarnekow: > wonder if we should add a testcase to
> AbstractXbaseEvaluationTest / XbaseInterpreterTest

Yes, that would be necessary. Xbase does not allow to define methods though the tests should be based on lambda expressions with a procedural type, e.g. `val Runnable r = [ return; ]; r.run()`

szarnekow: I don't think it's a regression. Looks like it was an incomplete solution since 2011

szarnekow: > an elseif for Null/Void in org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate

No, I don't think so. A null expression would be a programming error which is already indicated by the IllegalArgumentException.

szarnekow (commented):
@asheiduk The change looks good to me - thank you for providing the fix :+1:. Please add a test though as indicated by Christian.

asheiduk:
@szarnekow @cdietrich Sorry, I have to punt on the test -- that is beyond my actual Xtext-Internal level.

cdietrich:
i would exepct something simple like
```
	@Test public void testReturnExpression_Void() throws Exception {
		assertEvaluatesTo(null, "return");
	}
```
should show the problem and fix

szarnekow:
@asheiduk Do you want to give it a shot?

asheiduk:
@szarnekow I'd prefer to pass this one as already mentioned. 

I debugged and patched the source with "import plugin as source" (sic?) and it took some time to nail the thing. But setting up a complete developer environment to execute the unittests was a disaster and time-sink the last time I've tried. Sidenote: That last attempt already included Oomph. I'd like not to spend more time on that.

cdietrich (approved):

cdietrich:
will provide a test separately

cdietrich:
@asheiduk which os do you use to setup with oomph? windows?

asheiduk:
@cdietrich I tried that on Linux quite a while ago. We've talked about that at that time if I remember correctly.

PR #681, opened 2020-12-08
test for return; in xbase interpreter

cdietrich:
test for return; in xbase interpreter


cdietrich:
cc @asheiduk

szarnekow (approved):

PR #688, opened 2021-01-27
[eclipse/xtext#1923] Change visibility

miklossy:
[eclipse/xtext#1923] Change visibility

of the XbaseInterpreter.ReturnValue class from package private to
public.

Signed-off-by: miklossy <miklossy@itemis.de>

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2016 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2010, 2021 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
```diff
@@ -115,7 +115,7 @@
 public class XbaseInterpreter implements IExpressionInterpreter {
 	
 	@SuppressWarnings("serial")
-	static class ReturnValue extends RuntimeException {
+	public static class ReturnValue extends RuntimeException {
```
szarnekow, line 118: Can you please add a `@Since` tag for documentation purpose. If you don't mind, this class would now also deserve a brief JavaDoc comment to clarfiy its purpose.

szarnekow (approved):
Looks good - pending the missing since tag.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
miklossy: since tag has been added.

miklossy: JavaDoc comment has been added.

szarnekow (approved):
I find the part about the TryCatchFinally slightly misleading, but generally speaking, this is correct. Approved.

cdietrich:
@szarnekow could you help rephrasing? or should we leave it out?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java
```diff
@@ -114,8 +114,16 @@
  */
 public class XbaseInterpreter implements IExpressionInterpreter {
 	
+	/**
+	 * Encapsulates the return value of an {@link XReturnExpression}
```
szarnekow, line 118:
```suggestion
Encapsulates the return value of an {@link XReturnExpression}.

To be able to break up the control flow of the interpreter, this class inherits from {@link RuntimeException}.
Clients that catch the ReturnValue, should usually re-throw it further. See also {@link _doEvaluate(XTryCatchFinallyExpression, ..)} where the finally clause is evaluated after the return value has been computed but the original return value is propagated further, if the finally block completes normally.
```

I would be fine with keeping it concise and simply remove the part about the TryCatchFinallyExpression.

miklossy: Thanks for the explanation @szarnekow ! I updated the PR accordingly.

cdietrich (approved):

PR #692, opened 2021-02-12
[eclipse/xtext-core#1626] regnerate languages 

cdietrich:
[eclipse/xtext-core#1626] regnerate languages

### DIFF
--- a/org.eclipse.xtext.extras.tests/src-gen/org/eclipse/xtext/enumrules/services/EnumRulesTestLanguageGrammarAccess.java
+++ b/org.eclipse.xtext.extras.tests/src-gen/org/eclipse/xtext/enumrules/services/EnumRulesTestLanguageGrammarAccess.java
@@ -41,7 +41,8 @@ public class EnumRulesTestLanguageGrammarAccess extends AbstractElementFinder.Ab
 		private final RuleCall cGeneratedGeneratedEnumEnumRuleCall_1_1_0 = (RuleCall)cGeneratedAssignment_1_1.eContents().get(0);
 		
 		//Model:
-		//	'existing' existing=ExistingEnum ('generated' generated=GeneratedEnum)? | 'generated' generated=GeneratedEnum;
+		//  'existing' existing=ExistingEnum ('generated' generated=GeneratedEnum)? | 'generated' generated=GeneratedEnum
+		//;
 		@Override public ParserRule getRule() { return rule; }
 		
 		//'existing' existing=ExistingEnum ('generated' generated=GeneratedEnum)? | 'generated' generated=GeneratedEnum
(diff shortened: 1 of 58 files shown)


org.eclipse.xtext.extras.tests/src-gen/org/eclipse/xtext/enumrules/services/EnumRulesTestLanguageGrammarAccess.java
```diff
@@ -95,28 +96,27 @@
 		private final Keyword cDifferentNameDifferentLiteralKeyword_2_0 = (Keyword)cDifferentNameEnumLiteralDeclaration_2.eContents().get(0);
 		
 		//enum ExistingEnum:
-		//	SameName | OverriddenLiteral="overridden" | DifferentName="DifferentLiteral";
+		//  SameName | OverriddenLiteral = "overridden" | DifferentName
```
ArneDeutsch, line 99: Interesting ... the old logic using the serializer infers the value and generates an assignment, that is not present in the grammar. Of course the new logic is not capable of that.

org.eclipse.xtext.xbase/src-gen/org/eclipse/xtext/xbase/services/XbaseGrammarAccess.java
```diff
@@ -96,21 +99,23 @@
...
+		//    =>({XBinaryOperation.leftOperand=current} feature=[types::JvmIdentifiableElement|OpMultiAssign]) rightOperand=XAssignment
+		//)?
 		public Group getGroup_1() { return cGroup_1; }
 		
 		//XOrExpression
 		public RuleCall getXOrExpressionParserRuleCall_1_0() { return cXOrExpressionParserRuleCall_1_0; }
 		
-		//(=> ({XBinaryOperation.leftOperand=current} feature=[types::JvmIdentifiableElement|OpMultiAssign])
-		//rightOperand=XAssignment)?
+		//(
```
ArneDeutsch, line 110: Here is one of the rare cases that works not so well. But I would consider it minor.

ArneDeutsch (approved):

org.eclipse.xtext.xbase/src-gen/org/eclipse/xtext/xbase/services/XbaseGrammarAccess.java
szarnekow: Agreed

szarnekow (approved):

org.eclipse.xtext.extras.tests/src-gen/org/eclipse/xtext/enumrules/services/EnumRulesTestLanguageGrammarAccess.java
szarnekow: For completeness: The serializer operates on the model which is augmented by the Xtext linker. The linker will create objects like the return type ref or the string literal for enumerations if the short-hand notation is used.
The transient value service is later responsible for skipping these again - but it is not perfect: The string literal `DifferentLiteral` would be optional.

Issue #695, opened 2021-02-17
Inconsistent handling of synthetic fragment for missing obj 


dhuebner:
In `BatchLinkableResourceStorageWritable` a "none" fragment is used to probably mark a missing unresolved EObject. Which is fine I think:

https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/resource/BatchLinkableResourceStorageWritable.java#L158

Later on in `BatchLinkableResourceStorageLoadable` one try to resolve the EObject using the fragment "none" which leads to a null key entry in the `targetToSourceMap` 

https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/resource/BatchLinkableResourceStorageLoadable.java#L87

The bad thing about, is that the `BatchLinkableResourceStorageWritable` uses `targetToSourceMap` to write an association map and fails with an NullPointerException inside the `logIfResourceMismatch` because of the `null` key:
https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/resource/BatchLinkableResourceStorageWritable.java#L144

Because of that NPE an Xtend resource can not be loaded/indexed. 

The problem with this synthetic fragment string seem to be old and I'm not sure why it now breaks the build. Probably because of migration from Xtend -> Java and new try{}finally{} replacement. 

I see two possible actions:
1. Guard  `BatchLinkableResourceStorageWritable.logIfResourceMismatch` for `null` object
2. Don't produce `null` key by checking for the "none" fragment or better for the `getFragment(keyObject) == null` see  BatchLinkableResourceStorageLoadable.readAssociationsAdapter(BatchLinkableResource, ZipInputStream)

dhuebner:
@szarnekow 
Does it make sense?

szarnekow:
The analysis sounds about right to me. It would be great to have a reproducer.

dhuebner:
@szarnekow 
Create a reproducer is harder than I expected. It turned out that the EObject that I mentioned here:
>  a "none" fragment is used to probably mark a missing unresolved EObject. Which is fine I think:

... is a synthetic constructor for an Xtend class that is annotated with a `@FinalFieldsConstructor` annotation and the bogus case here is that this JvmConstructor doesn't have an EContainer :-/ 


> Which is fine I think

[...]

PR #698, opened 2021-02-25
Testcase for #695

dhuebner:
Test case for  Inconsistent handling of synthetic fragment for missing or orphaned obj #695 

The test adds a fake constructor to a JvmGenericType and adds associations.
After removing this constructor from the original resource the resource is stored again causing a "none" fragment created.

Loading the stored bad state causes `null` keys in `org.eclipse.xtext.xbase.jvmmodel.JvmModelAssociator.Adapter.targetToSourceMap`

Storing this state than cause an NPE in `org.eclipse.xtext.xbase.resource.BatchLinkableResourceStorageWritable.logIfResourceMismatch(BatchLinkableResource, EObject)`


org.eclipse.xtext.xbase.tests/src/org/eclipse/xtext/xbase/tests/resources/ResourceStorageTest.java
```diff
@@ -163,4 +170,47 @@ public void testWriteAndLoad() throws Exception {
...
+		type.getMembers().add(jvmConstructor);
+		associator.associate(file.eResource().getContents().get(0), jvmConstructor);
+
+		StorageAwareResource originalResource = ((StorageAwareResource) file.eResource());
+		ByteArrayOutputStream bout = new ByteArrayOutputStream();
+		resourceStorageFacade.setStoreNodeModel(false);
+		resourceStorageFacade.createResourceStorageWritable(bout).writeResource(originalResource);
+		
+		// Remove synthetic constructor to cause an addition of a "none" fragment
+		type.getMembers().remove(jvmConstructor);
```
szarnekow, line 195: @dhuebner I can understand that it'S possible to break the serialization if the model is corrupted. I'd be curious though how that state was "achieved" in the real case.

dhuebner: @szarnekow It happens often when switching a branch, sometimes it also happens during regular work/typing. What I see is that the broken Xtend file has an active annotation or an inner Xtend class has one. `@Data` or `@FinalFieldConstructor` for example manipulating constuctors of the JvmType, so I suppose it's a kind of race condition. The problem is that I can't see why the constructor eResource not in place when a `none` fragment was added.

dhuebner:
@szarnekow 
I prepared a fix for the "none" fragment case.
What would you think about having a logging for the bogus case and trigger further investigation in Xtend with an Active annotation context?

szarnekow:
Sounds good to me. The patch looks fine, too at first glance.

dhuebner:
@szarnekow 
I see a lot of errors logged in my workspace. As we don't know the cause yet and to not annoy the users, I changed log level to `debug`. I also created a new issue in xtext-xtend eclipse/xtext-xtend#1189

@cdietrich @szarnekow 
Could some of you approve this PR?

cdietrich:
TODO create/update downstream builds in eclipse and xtend to see what breaks

cdietrich:
pushed branches

dhuebner:
> TODO create/update downstream builds in eclipse and xtend to see what breaks

Is that a TODO for me?

cdietrich:
> Is that a TODO for me?
i have already done it. (in theory would be a todo for you)

dhuebner:
@cdietrich
>  i have already done it.

Thanks

cdietrich:
no test fails.

dhuebner:
@cdietrich 
I pushed a commit with error logging. I also noticed and fixed a potential NPE. 
... and I don't think it's a useful information that an eProxy doesn't have an eResource

cdietrich:
ok will restart the build once yours went through

dhuebner:
With a change above I see only few error logs. And that all synthetic `JvmConstructor` from Xtend classes that are annotated with `@Data` or `@FilnalFieldConstructor` and are inside an Xtend file but are not the primary Xtend types. Means this types share the same Xtend file with other Xtend types

dhuebner:
@cdietrich 
I see any log entires in jenkins console log

cdietrich:
me neither

dhuebner:
@cdietrich 
Can we merge it?

cdietrich:
i want to wait for @szarnekow am not sure if the error logging is working.

dhuebner:
In xbase the logging is working:
https://ci.eclipse.org/xtext/job/xtext-extras/job/dh-storage-writable-695/6/testReport/org.eclipse.xtext.xbase.tests.resources/ResourceStorageTest/testWriteAndLoadWithMissing/

szarnekow (approved):

szarnekow:
Looks about right to me.

PR #703, opened 2021-03-15
make tests runnable with java 15 and 16

cdietrich:
make tests runnable with java 15 and 16


cdietrich:
ReusedTypeProviderTest > testMemberCount_16 and other testMemberCount_16 are still failing.
Looks like Enums are compiled differently with Java 15
@szarnekow any idea?

szarnekow:
When we compile xbase.testdata with compliance level 11 or newer, there is no synthetic constructor in being generated.
Probably because nestmates allow to access the real stuff now without synthetic bridges.

cdietrich:
ok i propose to leave the tests as is right now and just to commit the fixes i have done

szarnekow (approved):
Agreed

PR #716, opened 2021-08-12
update to gradle 7

cdietrich:
update to gradle 7
See https://github.com/eclipse/xtext/issues/1548


gradle/longrunning-tests.gradle
```diff
@@ -38,6 +38,12 @@ jar {
 }
 
 sourcesJar {
+	if (findProperty('compileXtend') == 'true') {
+		// TODO conditional is this correct?!?
```
oehme, line 42: Probably the best we can do, given our check-in-sources approach.

gradle/mwe2-workflows.gradle
```diff
@@ -5,14 +5,14 @@
 sourceSets {
 	mwe2 {}
 }
-
-eclipse.classpath.plusConfigurations += [configurations.mwe2Runtime]
+configurations.mwe2Implementation.canBeResolved = true //TODO is there a better way
+eclipse.classpath.plusConfigurations += [configurations.mwe2Implementation]
```
oehme, line 9: Use mwe2RuntimeClasspath instead.

szarnekow (approved):

PR #722, opened 2021-08-19
[eclipse/xtext-extras#695] fix removeAssociation leaving orphans in target2source map

cdietrich:
[eclipse/xtext-extras#695] fix removeAssociation leaving orphans in target2source map

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2021 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
```diff
@@ -467,6 +467,9 @@ public void removeAssociation(EObject sourceElement, EObject jvmElement) {
 		if (sources != null && sources.remove(sourceElement)) {
 			Set<EObject> targets = sourceToTargetMap(resource).get(sourceElement);
 			targets.remove(jvmElement);
+			if (sources.isEmpty()) {
```
cdietrich, line 470: have no idea how to test this in
org.eclipse.xtext.xbase.tests.jvmmodel.JvmModelAssociatorTest.testRemoveAssociation()
as it always will return empty set.

cdietrich: am not sure about removeAssocian vs removeAllAssociation

cdietrich: testcase in ResourceStorageTest
does not call remove*Assoc at all.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
```diff
@@ -467,6 +467,9 @@ public void removeAssociation(EObject sourceElement, EObject jvmElement) {
 		if (sources != null && sources.remove(sourceElement)) {
```
szarnekow, line 467: Please make a local var for the `targetToSourceMap(resource)` and use that one here and below.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
```diff
@@ -467,6 +467,9 @@ public void removeAssociation(EObject sourceElement, EObject jvmElement) {
...
 			Set<EObject> targets = sourceToTargetMap(resource).get(sourceElement);
 			targets.remove(jvmElement);
```
szarnekow, line 469: What if targets is empty, now? Should it be removed from the sourceToTargetMap, too?

szarnekow (approved):

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmModelAssociator.java
cdietrich: did not check if that would have any side effects

szarnekow: Ok

Issue #723, opened 2021-08-19
testMemberCount_16 fails with java 17 (and maybe also 16)


cdietrich:
testMemberCount_16 fails with java 17 (and maybe also 16)
needs also to be tested with eclipse variant of AbstractTypeProviderTest

cdietrich:
done

PR #724, opened 2021-08-19
[#723] Fix TypeProvider expectations for Java 17 Enum style 

cdietrich:
[#723] Fix TypeProvider expectations for Java 17 Enum style

### DIFF
--- a/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
+++ b/org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2021 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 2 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.1.1-bin.zip
```
cdietrich, line 3: this change is from separate PR and needed to run with java 17

org.eclipse.xtext.common.types.tests/tests/org/eclipse/xtext/common/types/access/impl/AbstractTypeProviderTest.java
```diff
@@ -819,24 +819,46 @@ public void testMemberCount_15() {
...
-		assertEquals(2, constructorCount);
-		int fieldCount = TestEnum.class.getDeclaredFields().length;
-		// FirstValue, SecondValue, string, ENUM$VALUES
-		assertEquals(Arrays.toString(TestEnum.class.getDeclaredFields()), 4, fieldCount);
-		// ENUM$VALUES is synthetic
-		// TestEnum(String, String, EnumType) is synthetic
-		assertEquals(type.getMembers().toString(), innerTypesCount + methodCount + constructorCount + fieldCount - 2,
-				type.getMembers().size());
+		try {
+			String typeName = TestEnum.class.getName();
```
cdietrich, line 823: am not sure what our common pattern in extras is to deal with different java versions

szarnekow: I don't think we do have a pattern for that.
How did you like the assumeTrue(..) thingy?

cdietrich: we have the old variant. will apply it here

cdietrich: problem is eclipse. the enum is compiled differently there as of now

PR #769, opened 2022-04-01
[eclipse/xtext-xtend#1301] fix name clashes with anony. inner classes

cdietrich:
[eclipse/xtext-xtend#1301] fix name clashes with anony. inner classes

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2018 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2010, 2022 itemis AG (http://www.itemis.eu) and others.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -1912,7 +1912,7 @@ protected void appendClosureParameter(JvmFormalParameter closureParam, Lightweig
 		appendable.append(parameterType);
 		appendable.append(" ");
 		final String proposedParamName = makeJavaIdentifier(closureParam.getName());
-		String name = appendable.declareVariable(closureParam, proposedParamName);
+		String name = appendable.declareUniqueNameVariable(closureParam, proposedParamName);
```
cdietrich, line 1915: have no idea if this is the correct solution

szarnekow: It appears to be ok. In the compiled Java code, it'd be possible to refer to outer variables now which would be shadowed in Xbase itself, but that's not a problem.

szarnekow (approved):

PR #789, opened 2022-06-18
switch to java 8 so that ecj is running with annotation proccessing

cdietrich:
switch to java 8 so that ecj is running with annotation proccessing


szarnekow (approved):
Change makes sense but I don't quite get why it's not running with APT when this is not set? Does it assume java 1.3 or something?

cdietrich:
I think default is 1.5 or so. Annotation processing is a 1.6 feature @szarnekow should I keep 8 or go back to 1.6

szarnekow:
> I think default is 1.5 or so. Annotation processing is a 1.6 feature @szarnekow should I keep 8 or go back to 1.6

Java 8 is the way to go - soonish it'll be Java 11 anyway.

PR #795, opened 2022-08-12
change baseclass for xbase formatter

cdietrich:
change baseclass for xbase formatter


cdietrich:
@szarnekow  what do you think about this idea. might make creating java subclasses more easy.
am not sure if we also might want to change the dispatcher to fetch _format methods instead format methods

szarnekow:
> @szarnekow what do you think about this idea. might make creating java subclasses more easy. am not sure if we also might want to change the dispatcher to fetch _format methods instead format methods

I like the idea. Also the comments `org.eclipse.xtext.formatting2.AbstractJavaFormatter.format(Object, IFormattableDocument)` and in the classes JavaDoc suggest that it should be `_format` indeed. Shall we find both names?

cdietrich:
unfortunately using _format and no dispatch does not seem to work
All local dispatch methods must have the same visibility.

cdietrich:
we could do the follwing

in AbstractJavaFormatter
```
@Override
	public void format(Object child, IFormattableDocument document) {
		formatUsingPolymorphicDispatcher(child, document);
	}

	/**
	 * since
	 */
	protected void formatUsingPolymorphicDispatcher(Object child, IFormattableDocument document) {
		if (child instanceof XtextResource) {
			_format((XtextResource) child, document);
			return;
		} else if (child == null) {
			_format((Void) null, document);
			return;
		}
		try {
			dispatcher.invoke(child, document);
		} catch (SecurityException | IllegalArgumentException e) {
			throw new RuntimeException(e);
		}
	}
```

in XtypeFormatter:
```
	override protected PolymorphicDispatcher<Void> createPolymorhicDispatcher() {
		return new PolymorphicDispatcher<Void>(Collections.singletonList(this), [ m |
			"_format".equals(m.getName()) && m.getParameterCount() === 2 && m.getParameterTypes().get(0) !== Object &&
				m.getParameterTypes().get(1) === IFormattableDocument
		], new EObjectErrorHandler())
	}
```

in JavaSubclasses (needs to be done manually)
```
	@Override
	public void format(Object expr, IFormattableDocument format) {
		formatUsingPolymorphicDispatcher(expr, format);
	}
```

cdietrich:
should i update docs of the baseclass?

szarnekow (approved):

PR #808, opened 2022-09-28
Fixed some warnings emitted during gradle build

szarnekow:


org.eclipse.xtext.common.types/emf-gen/org/eclipse/xtext/common/types/JvmTypeConstraint.java
```diff
@@ -82,7 +82,7 @@ public interface JvmTypeConstraint extends EObject
 	 * <p>The identifier of a JvmTypeConstraint is a canonical representation of the constraint information.</p>
 	 * <p>Examples for constraint identifiers are:</p>
 	 * <ul>
-	 * <li>extends java.util.List<java.lang.Map$Entry<java.lang.Object,java.lang.Object>></li>
```
cdietrich, line 85: these will do when we regenerate the model right?

cdietrich: is this fixed in a newer emf version?

szarnekow: I patched the JavaVMTypes.ecore and replaced to bogus chars in the documentation annotations.
Regenerating the model code from there fixed the Java code.

cdietrich (approved):

Issue #815, opened 2022-10-01
StandaloneBuilder leads to test error in xtext-eclipse with tycho 2x.


cdietrich:
```
07:24:19  	[/opt/tools/java/temurin/jdk-11/latest/bin/java, -Dosgi.noShutdown=false, -Dosgi.os=linux, -Dosgi.ws=gtk, -Dosgi.arch=x86_64, -Xmx1500m, -Dosgi.clean=true, -jar, /home/jenkins/agent/workspace/xtext-eclipse_master/$VALUE/p2/osgi/bundle/org.eclipse.equinox.launcher/1.6.400.v20210924-0641/org.eclipse.equinox.launcher-1.6.400.v20210924-0641.jar, -data, /home/jenkins/agent/workspace/xtext-eclipse_master/org.eclipse.xtext.builder.tests/target/work/data, -install, /home/jenkins/agent/workspace/xtext-eclipse_master/org.eclipse.xtext.builder.tests/target/work, -configuration, /home/jenkins/agent/workspace/xtext-eclipse_master/org.eclipse.xtext.builder.tests/target/work/configuration, -application, org.eclipse.tycho.surefire.osgibooter.uitest, -testproperties, /home/jenkins/agent/workspace/xtext-eclipse_master/org.eclipse.xtext.builder.tests/target/surefire.properties, -pluginCustomization, /home/jenkins/agent/workspace/xtext-eclipse_master/org.eclipse.xtext.builder.tests/../releng/org.eclipse.xtext.tycho.tests.parent/pluginCustomization.ini, -timeout, 2400000]
07:24:31  
07:24:31  Results:
07:24:31  
07:24:31  Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
07:24:31  
07:24:31  java.lang.reflect.InvocationTargetException
07:24:31  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
07:24:31  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
07:24:31  	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
07:24:31  	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
07:24:31  	at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray2(ReflectionUtils.java:206)
[...]

Issue #814, opened 2022-10-01
StandaloneBuilderTest flaky/does not work on Linux


cdietrich:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.xtext.builder.standalone.StandaloneBuilderTest.testWriteStorageResource(StandaloneBuilderTest.java:90)
```
```
0    [main] INFO  uilder.standalone.StandaloneBuilder  - Collecting source models.
178  [main] INFO  uilder.standalone.StandaloneBuilder  - Validate and generate.
184  [main] INFO  uilder.standalone.StandaloneBuilder  - Validating: 'Bar.buildertestlanguage'
266  [main] INFO  uilder.standalone.StandaloneBuilder  - Validating: 'Foo.buildertestlanguage'
267  [main] INFO  uilder.standalone.StandaloneBuilder  - Starting generator for input: 'Bar.buildertestlanguage'
308  [main] INFO  uilder.standalone.StandaloneBuilder  - Starting generator for input: 'Foo.buildertestlanguage'
311  [main] INFO  uilder.standalone.StandaloneBuilder  - Build took 313ms.

```

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.xtext.builder.standalone.StandaloneBuilderTest.testDifferentOutputFolders(StandaloneBuilderTest.java:73)
```
```
0    [main] INFO  uilder.standalone.StandaloneBuilder  - Collecting source models.
171  [main] INFO  uilder.standalone.StandaloneBuilder  - Validate and generate.
176  [main] INFO  uilder.standalone.StandaloneBuilder  - Validating: 'Bar.buildertestlanguage'
243  [main] INFO  uilder.standalone.StandaloneBuilder  - Validating: 'Foo.buildertestlanguage'
244  [main] INFO  uilder.standalone.StandaloneBuilder  - Starting generator for input: 'Bar.buildertestlanguage'
258  [main] INFO  uilder.standalone.StandaloneBuilder  - Starting generator for input: 'Foo.buildertestlanguage'
259  [main] INFO  uilder.standalone.StandaloneBuilder  - Build took 261ms.

```

[...]

PR #816, opened 2022-10-01
Fix Standalone Builder bugs

cdietrich:
[#814] null current source folder for default source folder
[#815] fix MANIFEST.MF and align core.runtime and equinox.common deps with gradle


szarnekow (approved):
Thank you

org.eclipse.xtext.builder.standalone/src/org/eclipse/xtext/builder/standalone/StandaloneBuilder.java
```diff
@@ -694,6 +694,7 @@ protected void registerCurrentSource(URI uri) {
 					"Resource " + uri + " is not contained in any of the known source folders " + sourceDirs + ".");
 		}
 		URI projectBaseURI = UriUtil.createFolderURI(new File(baseDir));
+		fsa.setCurrentSource(null);
```
cdietrich, line 697: why does this break maven tests?

PR #831, opened 2022-11-14
provide a default implementation using ENamedElement.getName

LorenzoBettini:
This way, one can simply redefine the default implementation instead of single cases.

### DIFF
--- a/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
+++ b/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
@@ -12,6 +12,7 @@ import java.util.Collections;
 
 import org.eclipse.emf.ecore.EClassifier;
 import org.eclipse.emf.ecore.EEnumLiteral;
+import org.eclipse.emf.ecore.ENamedElement;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EOperation;
 import org.eclipse.emf.ecore.EPackage;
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
@cdietrich how to run tests of projects like `org.eclipse.xtext.extras.tests`? From Eclipse running them as JUnit or Plug-in JUnit tests make most of them fail because they don't seem to find resources...

cdietrich:
you run them as pure junit tests
and you can also run the
1-xxx and 3-xxx bash scripts in the project root

cdietrich:
can you give an example for a failing test?

LorenzoBettini:
@cdietrich then I guess we're talking about the same tests (isn't there any way to run such tests from the IDE?):

image

cdietrich:
no there is no way.
i did not find a solution that works on both eclipse and gradle and gave up some day

LorenzoBettini:
@cdietrich do you think we can merge this?

cdietrich:
@LorenzoBettini as we are beyond m3 i plan to have this in 2.30

org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
```diff
@@ -74,22 +75,31 @@ protected String getCacheKey() {
...
 
 	protected String name(EOperation eOperation) {
-		return eOperation.getName();
+		return defaultName(eOperation);
+	}
+
+	/**
+	 * Default implementation: simply returns {@link ENamedElement#getName()}.
+	 * 
+	 * @since 2.29
```
cdietrich, line 100: @LorenzoBettini can you change this one to 2.30. then we can merge

LorenzoBettini:
@cdietrich I rebased and changed the since

cdietrich:
lets see if build works again once the infrastructure problems on foundation side are resolved

szarnekow (approved):
LGTM

PR #833, opened 2022-11-14
Replace polymorphic dispatcher with EcoreSwitch

LorenzoBettini:
DON'T MERGE YET!

That's just a proposal (it introduces also a
org.eclipse.xtext.ecore.EcoreQualifiedNameProvider.defaultName(ENamedElement) as in https://github.com/eclipse/xtext-extras/pull/831, and this will probably be a conflict that should be resolved in case).

If I understand correctly, EMF switches are mostly constant time because they perform a switch with integers.

Could it be that using the EcoreSwitch might be faster than a polymorphic dispatch?

@cdietrich @szarnekow what do you think?
I wouldn't know how to measure the difference (I mean in our code base)
@merks is it true that the EMF switch aims at being constant time method dispatcher?

### DIFF
--- a/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
+++ b/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
@@ -8,18 +8,17 @@
  *******************************************************************************/
 package org.eclipse.xtext.ecore;
 
-import java.util.Collections;
-
 import org.eclipse.emf.ecore.EClassifier;
 import org.eclipse.emf.ecore.EEnumLiteral;
+import org.eclipse.emf.ecore.ENamedElement;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EOperation;
 import org.eclipse.emf.ecore.EPackage;
 import org.eclipse.emf.ecore.EStructuralFeature;
+import org.eclipse.emf.ecore.util.EcoreSwitch;
 import org.eclipse.xtext.naming.IQualifiedNameProvider;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.util.IResourceScopeCache;
-import org.eclipse.xtext.util.PolymorphicDispatcher;
 import org.eclipse.xtext.util.Strings;
 import org.eclipse.xtext.util.Tuples;
 
@@ -31,14 +30,6 @@ import com.google.inject.Provider;
  */
 public class EcoreQualifiedNameProvider extends IQualifiedNameProvider.AbstractImpl {
 
-	private PolymorphicDispatcher<String> nameDispatcher = new PolymorphicDispatcher<String>("name", 1, 1,
-			Collections.singletonList(this), PolymorphicDispatcher.NullErrorHandler.<String> get()) {
-		@Override
-		protected String handleNoSuchMethod(Object... params) {
-			return null;
-		}
-	};
-
 	@Inject
 	private IResourceScopeCache cache = IResourceScopeCache.NullImpl.INSTANCE;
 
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
> Yes, the generate EMF switch will get to the deepest type's caseAbc first in constant time, but it's a per-EPackage things. The polymorphic dispatch is like a long series of instance of tests and it works for any types so is nicer notation. For this case I do expect it to be significantly faster.

Thanks for response, @merks ! I forgot to mention that yes, my idea was to use EMF switch when the dispatch has to be done according EMF classes (of the same known EPackage), like this case (and, I guess, many other places in Xbase where we have a cascade of if instance checks and downcasts (linear time) that are based on EMF classes only, e.g., the type computer to mention one)

LorenzoBettini:
@merks I can only check the total time of the Jenkins jobs.

With the EcoreSwitch https://ci.eclipse.org/xtext/job/xtext-extras/job/lb_proposal_ecore_switch/ is slightly slower than without https://ci.eclipse.org/xtext/job/xtext-extras/job/lb_defaultName/ but I strongly doubt this can reliably say something about the efficiency ;)

org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreQualifiedNameProvider.java
```diff
@@ -49,7 +40,28 @@ public QualifiedName getFullyQualifiedName(final EObject obj) {
 			@Override
 			public QualifiedName get() {
 				EObject temp = obj;
-				String name = nameDispatcher.invoke(temp);
+				String name = new EcoreSwitch<String>() {
```
szarnekow, line 43: This might even be slower than before since it'll call from the most special type up to the default case for all objects that are not in here. Did you measure this e.g. with JMH?

Since it is an anonymous inner class, I don't see how this could be of real benefit. It is not really reusable.

If it is provably faster, I'd rather make EcoreQualifiedNameProvider an `@Singleton` and inherit from the EcoreSwitch with more elaborate overrides for the objects that are often present in an ePackage but don't have a name.

LorenzoBettini: > This might even be slower than before since it'll call from the most special type up to the default case for all objects that are not in here. Did you measure this e.g. with JMH?
> 

I never used JMH so I wouldn't know where to start from (but I'd like to learn of course, if you have some suggestions). Concerning the first part, do you mean that it's best to have cases for EClass, EDataType instead of EClassifier?

> Since it is an anonymous inner class, I don't see how this could be of real benefit. It is not really reusable.

Something like the private field as the polymorphic dispatcher used to be?

> 
> If it is provably faster, I'd rather make EcoreQualifiedNameProvider an `@Singleton` and inherit from the EcoreSwitch with more elaborate overrides for the objects that are often present in an ePackage but don't have a name.

This last part is not clear to me :)

szarnekow: Being unclear is hobby of mine I'm afraid :)

I was thinking
```java
@Singleton
public class EcoreQualifiedNameProvider extends EcoreSwitch<QualifiedName> implements IQualifiedNameProvider {
  // no caching no nothing, just fairly optimized computation of the QualifiedName for few types that are relevant
  // e.g.
  @Override
  public QualifiedName caseEStructuralFeature(EStructuralFeature object) {
    // might even go fancier here with a QualifiedName.builder(..)
    return doSwitch(object.getEContainingClass()).append(object.getName());
  }
  ..
}
```

cdietrich:
this repository has been merged/moved to https://github.com/eclipse/xtext.
please feel free to recreate this PR against that repository in case it still is relevant.

PR #840, opened 2022-12-03
mark old generator forRemoval

cdietrich:
mark old generator forRemoval


cdietrich:
@szarnekow what do you think?

szarnekow (approved):
Fine for me. Since it was already deprecated: should we „just“ remove it?

PR #852, opened 2022-12-16
Add description to build.gradle

szarnekow:


org.eclipse.xtext.purexbase/build.gradle
```diff
@@ -1,6 +1,7 @@
 apply from: "${rootDir}/gradle/mwe2-workflows.gradle"
 
 ext.title = 'Pure Xbase Runtime'
+description = 'Pure Xbase Runtime.'
```
cdietrich, line 4: inconsistent use of full stop

cdietrich: usual converntion seems to include the full stop so the ide one needs update.

cdietrich (approved):

PR #859, opened 2023-01-22
Move to Maven/Tycho for projects and build infrastructure

LorenzoBettini:
See https://github.com/eclipse/xtext-extras/issues/858

Later (maybe tomorrow) I'll describe the main changes.

Jenkins also builds fine.

PR #862, opened 2023-02-07
Lb 2052 cherry pick

LorenzoBettini:
This uses some changes of https://github.com/eclipse/xtext-extras/pull/859 see also https://github.com/eclipse/xtext/issues/2135#issuecomment-1415343116


LorenzoBettini:
@szarnekow looks like the commit https://github.com/eclipse/xtext-extras/pull/862/commits/050d6167e29f00eed499fd24c0712125e21298ca adapted from my main PR does not work when run from Gradle. I'd ever that in this PR context, because those changes are based on the fact that it should run in an OSGI environment, OK?

cdietrich (approved):

PR #863, opened 2023-02-08
also drop simpleJar from gradle

cdietrich:
also drop simpleJar from gradle


LorenzoBettini (approved):
That's consistent with https://github.com/eclipse/xtext-extras/commit/c5c43ebe44648b6fcde3ad7f62fca0b606f42556

PR #865, opened 2023-02-17
[eclipse/xtext-xtend#149] allow extension fields in java 8 interfaces

cdietrich:
[eclipse/xtext#2259] allow extension fields in java 8 interfaces

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/LogicalContainerAwareReentrantTypeResolver.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/LogicalContainerAwareReentrantTypeResolver.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2012, 2023 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/LogicalContainerAwareReentrantTypeResolver.java
```diff
@@ -1011,10 +1011,12 @@ protected IFeatureScopeSession addExtensionFieldsToMemberSession(
 				}
 			}
 			// traverse the type hierarchy to create the feature scope sessions
-			JvmTypeReference superType = getExtendedClass(type);
 			IFeatureScopeSession result = featureScopeSession;
-			if (superType != null) {
-				result = addExtensionFieldsToMemberSession(resolvedTypes, featureScopeSession, (JvmDeclaredType) superType.getType(), thisFeature, seenNames, seenTypes);
+			for (JvmTypeReference ifx : type.getSuperTypes()) {
```
cdietrich, line 1015: is there a better way? do we need sorting?

szarnekow: Yes, we'll need some sorting here to get a stable or understandable precedence. Now that we have to consider a graph due to multiple interface inheritance and also need to support duplicates that are *allowed* in that graph (following the superclass alone would only ever see dups for invalid hierarchies), a predictable order must be used.

org.eclipse.xtext.xbase.typesystem.override.ResolvedFeatures.computeAllOperationsFromSortedSuperTypes(JvmDeclaredType, Multimap<String, AbstractResolvedOperation>) might be a good starting point.

cdietrich: arggg guess ill postpone until i find time to digg into

cdietrich:
replaced with https://github.com/eclipse/xtext/pull/2591

